PR #1, opened 2020-08-30
Makefile changes: Remove objcopy hack and allow building with WSL

Thar0:
- Identifies WSL as windows since exe just works™️ on WSL
 - Removes the objcopy hack when building the elf after a change to elf2dol that makes it unnecessary
 - Changes the mwcc version to 3.0 to avoid the linker error occurring on linker versions below 2.7


Pheenoh:
👍

PR #4, opened 2020-11-29
Adds JSUList, JKRDisposer, and JKRHeap.

Julgodis:
Adds functions for the three following classes: JSUList, JKRDisposer, and JKRHeap. Most of the code changes are from renaming function labels to their mangled function name. 

Changes:
* **JSupport/JSUList** - Decompiled JSUPtrLink and JSUPtrList.. The templated class JSUList will have functions for each instatiation and will be in different translation units.  
* **JKernel/JKRDisposer** - Constructor matched but not the destructor.
* **JKernel/JKRHeap** - Class definition and ASM functions. JKRHeap.cpp now includes _operator delete_ which makes it possible to forward compile destructors.


ProjectRevoTPP (approved):

Pheenoh:
Nice work! Can you move the JS* and JK* translation units to the folders setup in `libs/` instead of in `src/` ? I had setup `src/` to be anything that wasn't in an archive file.

PR #16, opened 2020-12-05
f_pc_layer

lepelog:
some structs are definetly wrong, no idea how proper structs for the iter looks like
the makefile change made it work on linux (compiling crashed otherwise) but probably not the best solution
some dumb edits to structs were needed to get the struct copies to match, idk if there are better ways for that


include/f/f_pc/f_pc_layer_tag.h
```diff
@@ -7,10 +7,18 @@
 
 struct layer_class;
 
-struct layer_management_tag_class : public create_tag_class {
+struct layer_management_tag_class {
+    create_tag_class mCreateTag;
```
magcius, line 11: are you sure this isn't inheritance?

src/f/f_op/f_op_actor_tag.cpp
```diff
@@ -1 +1,21 @@
-// ok
+#include "f/f_op/f_op_actor_tag.h"
+#include "global.h"
+#include "SComponent/c_tag.h"
+
+extern node_list_class lbl_803A35E0;
```
magcius, line 5: please comment what the global is called so we can rename it later

src/f/f_pc/f_pc_layer.cpp
```diff
@@ -1 +1,149 @@
...
+    void fpcLy_Create(layer_class *pLayer, process_node_class *pPcNode, node_list_class *pLists, int listNum) {
+        void *pvVar1;
+        int iVar2;
+        
+        if (lbl_80450D20 == 0) {
+            lbl_80450D1C = 0x1; // IsInitOfLayerList
+            lbl_80450D20 = 1;
+        }
+        if (lbl_80450D28 == 0) {
+            lbl_80450D24 = 0; // layer_id
```
magcius, line 131: these are originally function statics, but I don't know if we can correctly lay them out right now

src/f/f_pc/f_pc_layer.cpp
```diff
@@ -1 +1,149 @@
...
+extern layer_class *lbl_80450D18;
+extern s8 lbl_80450D20;
+extern s32 lbl_80450D1C;
+extern s8 lbl_80450D28;
+extern s32 lbl_80450D24;
+extern node_list_class lbl_803A39DC;
+
+extern "C" {
+
+    void fpcLy_CancelQTo(process_method_tag_class *pMthd) {
```
magcius, line 15: please don't indent inside extern "C"

include/f/f_pc/f_pc_layer_iter.h
```diff
@@ -0,0 +1,20 @@
+#ifndef F_PC_LAYER_ITER_H
+#define F_PC_LAYER_ITER_H
+
+#include "global.h"
+#include "f/f_pc/f_pc_layer.h"
+#include "SComponent/c_node_iter.h"
+
+struct layer_iter {
+    void* pFunc;
+    void* pUserData;
```
magcius, line 10: mpFunc / mpUserData for struct members

include/SComponent/c_tag.h
```diff
@@ -7,7 +7,8 @@
 #include "SComponent/c_list.h"
 #include "SComponent/c_tree.h"
 
-struct create_tag_class : public node_class {
+struct create_tag_class {
+    node_class pNode;
```
magcius, line 11: again, i think this is inheritance.

src/f/f_pc/f_pc_layer_tag.cpp
```diff
@@ -1 +1,66 @@
...
+                return 1;
+            }
+            return 0;
+        }
+    }
+
+    s32 fpcLyTg_QueueTo(layer_management_tag_class *pTag) {
+        if (fpcLy_QueueTo(pTag->mpLayer,&pTag->mCreateTag) == 1) {
+            pTag->mpLayer = NULL;
+            pTag->mNodeListID = -1; // only matches when unsigned
```
magcius, line 33: did you try 0xFFFF?

include/f/f_pc/f_pc_layer_tag.h
lepelog: Inheritance messes up the struct copy in `fpcLyTg_Init`, but I don't know much about C(++), so there is probably something I'm missing

Pheenoh (approved):

PR #18, opened 2020-12-05
d_item

TakaRikka:
most of d_item done + renamed gameinfo label


src/d/d_item/d_item.cpp
```diff
@@ -1 +1,1759 @@
...
+void item_func_S_MAGIC(void){
+    g_dComIfG_gameInfo.give_item.magic += 4;
+}
+
+void item_func_L_MAGIC(void){
+    g_dComIfG_gameInfo.give_item.magic += 8;
+}
+
+void item_func_BOMB_5(void){
+    addBombCount(112, 5);
```
Pheenoh, line 56: The 112, 113, etc. are all item ids. I defined a enum in d_save.h so you can use a more friendly name. This happens everywhere in this TU looks like, so let's go ahead and swap those in

src/d/d_item/d_item.cpp
```diff
@@ -1 +1,1759 @@
...
+
+void item_getcheck_func_POKE_BOMB(void){
+    g_dComIfG_gameInfo.save_file.getPlayer().getPlayerGetItem().isFirstBit(114);
+}
+
+u32 item_getcheck_func_FAIRY_DROP(void){
+    u32 bottle_check;
+
+    bottle_check = g_dComIfG_gameInfo.save_file.getPlayer().getPlayerItem().checkBottle(115);
+    return bottle_check & 0xff;
```
Pheenoh, line 1320: These all may have been like this:

```
u8 item_getcheck_func_FAIRY_DROP(void){
    u8 bottle_check;

    bottle_check = g_dComIfG_gameInfo.save_file.getPlayer().getPlayerItem().checkBottle(115);
    return bottle_check;
}
```

and Ghidra just got the return type wrong (happens alot). Lmk!

Pheenoh (approved):

PR #27, opened 2020-12-10
add clang-format

notyourav:


lepelog:
Would it make sense to include a bash script for running the clang-format command?

notyourav:
> Would it make sense to include a bash script for running the clang-format command?

Shouldn't be necessary if it is part of CI check.

Pheenoh:
Added clang-format file + CI check in different commit

PR #29, opened 2020-12-12
m_Do_audio split and partially matched

jdflyer:


src/m_Do/m_Do_audio.cpp
```diff
@@ -1 +1,93 @@
...
+        if((int)lbl_80450C78->unk_4 == 0 && (int)lbl_80450C78->unk_5 == 0){
+            mDoAud_Create();
+        }
+    }else{
+        lbl_803DBF4C.gframeProcess();
+    }
+}
+
+void mDoAud_setSceneName(const char *spot,long room,long layer){
+    mAudioMgrPtr->sceneMgr.setSceneName((char*)spot, room, layer);
```
Pheenoh, line 26: Was the cast here actually needed? 🤔

src/m_Do/m_Do_audio.cpp
```diff
@@ -1 +1,93 @@
...
+    #include "m_Do/m_Do_audio/asm/mDoAud_Create.s"
+}
+
+void mDoAud_Execute(void) {
+    if (lbl_80450BB8 == 0) {
+        if((int)lbl_80450C78->unk_4 == 0 && (int)lbl_80450C78->unk_5 == 0){
+            mDoAud_Create();
+        }
+    }else{
+        lbl_803DBF4C.gframeProcess();
```
Pheenoh, line 21: Same thing here with the label rename. This is the global `g_mDoAud_zelAudio`

src/m_Do/m_Do_audio.cpp
jdflyer: Yeah, the function signature of setSceneName is just char* instead of const char*, and the signature of mDoAud_setSceneName is const char*, so a cast is needed to convert it.

include/Z2AudioLib/Z2AudioMgr/Z2AudioMgr.h
```diff
@@ -0,0 +1,15 @@
+#include "global.h"
```
lepelog, line 1:
```suggestion
#ifndef __Z2_AUDIO_MGR_H_
#define __Z2_AUDIO_MGR_H_
#include "global.h"
```

missing include guard

include/m_Do/m_Do_audio/m_Do_audio.h
```diff
@@ -0,0 +1,54 @@
+#include "global.h"
```
lepelog, line 1:
```suggestion
#ifndef __M_DO_AUDIO_H_
#define __M_DO_AUDIO_H_
#include "global.h"
```

PR #30, opened 2020-12-12
d_lib split and partially matched

jdflyer:


src/d/d_lib.cpp
```diff
@@ -1 +1,197 @@
...
+}
+#else
+asm void STControl::Yinit(void){
+    nofralloc
+    #include "d/d_lib/asm/STControl_NS_Yinit.s"
+}
+#endif
+
+double STControl::getValueStick(){
+    return lbl_803DD2E8.stickValue;
```
Pheenoh, line 65: Just a FYI you can correct the label in .bss to be more descriptive. Something like this:

```asm
.global m_cpadInfo
m_cpadInfo:
	.skip 0x100
```

which then would change these:

```cpp
double STControl::getValueStick() {
    return m_cpadInfo.stickValue;
}
```

etc etc

Pheenoh (approved):

PR #42, opened 2021-01-01
decomp some small files, do some cleanup

barzamin:
this pr is a bit of a random combination of various things i've been working on/cleaning up idly by looking at the smallest remaining `.s` files


src/d/d_save/d_save_init.cpp
```diff
@@ -1 +1,5 @@
-// ok
+#include "d/d_save/d_save_init/d_save_init.h"
+
+void setInitEventBit() {
+    return;
+}
```
notyourav, line 5: FYI, ending on anything but a newline violates C standard. A lot of text editors will add them back anyway.

notyourav (approved):

src/d/d_save/d_save_init.cpp
barzamin: i am not sure why my editor didn't add a final \n. i feel like clang-format should've regardless! anyway, i can fix it by hand if it's an issue

notyourav: No need, just wanted to mention in case you were trying to delete them manually.

barzamin:
@Pheenoh fixed merge, sorry for the misc PR with lots of conflict points 😅

PR #46, opened 2021-01-01
d_item getcheck improvements

TakaRikka:
fix getcheck funcs to be more accurate


notyourav (commented):
Lgtm, I'm assuming the ones you kept as `int` returns were non matching with `bool`?

TakaRikka:
yea, adds extra wrong instructions

notyourav (approved):

PR #47, opened 2021-01-02
JKRAssertHeap OK

Julgodis:


libs/JSystem/JKernel/JKRAssertHeap.cpp
```diff
@@ -1,102 +1,105 @@
 #include "JSystem/JKernel/JKRAssertHeap/JKRAssertHeap.h"
 #include "global.h"
 
-asm JKRAssertHeap::JKRAssertHeap(void*, unsigned long, JKRHeap*, bool) {
-    nofralloc
-#include "JSystem/JKernel/JKRAssertHeap/asm/func_802D12C4.s"
-}
+// #include "JSystem/JKernel/JKRAssertHeap/asm/func_802D12C4.s"
```
notyourav, line 4: If the function is decompiled and matching, we no longer need the assembly. You should be deleting the relevant assembly files.

notyourav (changes requested):

Pheenoh (approved):

PR #48, opened 2021-01-02
OSX support

notyourav:
Tested on WSL and PowerShell. Had to add a step for creating the build directories, which is the only thing that might break.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 WINDOWS := $(shell which wine ; echo $$?)
+UNAME_S := $(shell uname -s)
 
 #-------------------------------------------------------------------------------
 # Files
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -41,13 +42,20 @@ else
   WINE := wine
 endif
 
+# Hack for OSX
+ifeq ($(UNAME_S),Darwin)
+	CPP     := cpp-10 -P
```
barzamin, line 47: is `cpp-10` tied to a specific macos version

Makefile
```diff
@@ -79,12 +87,13 @@ SBSS_PDHR := 10
...
 default: all
 
-all: $(DOL)
+all: dirs $(DOL)
 
 ALL_DIRS := build $(BUILD_DIR) $(addprefix $(BUILD_DIR)/,$(SRC_DIRS) $(ASM_DIRS))
 
 # Make sure build directory exists before compiling anything
-DUMMY != mkdir -p $(ALL_DIRS)
+dirs:
```
barzamin, line 95: add `dirs` to `.PHONY`

barzamin (approved):
🍎

Makefile
notyourav: No, but it is the only alternative I could come up with. Apple decided to claim `gcc`, `g++`, `cpp` and others as aliases to their modified clang.

Pheenoh (approved):

PR #54, opened 2021-01-02
Fixed _savegpr etc.

Julgodis:
Renamed all func_XXX to _savegpr, _restgpr, _savefpr, and _restfpr.


include/functions.h
```diff
@@ -20,11 +20,11 @@ void PSMTXScale(void);
 void screenSet__12dFile_info_cFv(void);
 void dFile_info_c_NS_screenSet(void);
 void mDoExt_removeMesgFont(void);
-void func_80362210(void);
```
barzamin, line 23: ideally we should deduplicate and group `_savegpr` and `_restgpr` functions in this file; perhaps consider a `clib.h` included into `global.h`

barzamin (approved):

include/functions.h
notyourav: @barzamin They are compiler intrinsics, I think it would make more sense to keep them with MWCC related stuff.

notyourav (approved):

include/functions.h
Julgodis: @barzamin @notyourav Moved _savegpr and _restgpr functions to a new file "mwcc.h"

Pheenoh:
Renamed some (but not all) of these in my latest push as well. Feel free to chuck those changes in favor of the mwcc.h when you fix conflicts

Pheenoh (approved):

PR #55, opened 2021-01-02
diff.py: be smarter about finding object files

barzamin:


Pheenoh:
Fixed the expected bug in my last push, but this is a better idea. Feel free to overwrite those changes with this

Pheenoh (approved):

PR #57, opened 2021-01-03
TObject and JStudio internals

notyourav:
no longer a draft


include/JSystem/JGadget/binary.h
```diff
@@ -0,0 +1,37 @@
...
+#define JSYSTEM_JGADGET_BINARY_H
+
+#include "dolphin/types.h"
+
+namespace JGadget {
+namespace binary {
+
+inline u32 align_roundUp(u32 arg0, u32 uAlign) {
+    return (arg0 + uAlign - 1) & ~(uAlign - 1);
+}
```
Julgodis, line 11: There is already alignment macros in global.h, but maybe functions are more nice. Either way, it's probably the best to not have multiple ways to do the same thing.

notyourav: This is an inline from the debug version.

Pheenoh (approved):

PR #59, opened 2021-01-03
JKernel cleanup and DVD types

Julgodis:
Updated old JKernel code to follow the code style:
* removed this->
* added const
* added field offsets
* fixed naming
* removed // #include "XXX.s" for matched functions
* DVDFileInfo and DVDCommandBlock

### DIFF
--- a/libs/JSystem/JKernel/JKRDvdFile.cpp
+++ b/libs/JSystem/JKernel/JKRDvdFile.cpp
@@ -85,8 +85,7 @@ void JKRDvdFile::close() {
 }
 
 s32 JKRDvdFile::readData(void* param_1, long length, long param_3) {
-    // debug version have this:
-    // ASSERT((length & 0x1f) == 0);
+    JUT_ASSERT((length & 0x1f) == 0);
 
     OSLockMutex(&mMutex1);
     if (mOSThread) {
--- a/libs/JSystem/JKernel/JKRDvdFile.cpp
+++ b/libs/JSystem/JKernel/JKRDvdFile.cpp
@@ -32,16 +32,9 @@ asm JKRDvdFile::JKRDvdFile(long) {
 }
 #endif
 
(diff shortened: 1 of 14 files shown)


libs/JSystem/JKernel/JKRDvdFile.cpp
```diff
@@ -33,7 +34,7 @@ asm JKRDvdFile::JKRDvdFile(long) {
 
 #ifndef NONMATCHING
```
notyourav, line 35:
```
#ifndef NONMATCHING
```
This will resolve to choosing the non asm function when NONMATCHING is not defined, is there a reason we need the asm at all now?

libs/JSystem/JKernel/JKRDvdFile.cpp
```diff
@@ -94,56 +91,50 @@ void JKRDvdFile::close() {
     }
 }
 
-#define JUT_ASSERT(CONDITION)
-
-//#include "JSystem/JKernel/JKRDvdFile/asm/func_802D99B4.s"
 s32 JKRDvdFile::readData(void* param_1, long length, long param_3) {
-    JUT_ASSERT((length & 0x1f) == 0);
+    // debug version have this:
+    // ASSERT((length & 0x1f) == 0);
```
notyourav, line 96: I think it would be better (if we are keeping track of debug asserts) to use a nulled macro as opposed to comments.

Also, macro should be moved.

notyourav (changes requested):

libs/JSystem/JKernel/JKRDvdFile.cpp
Julgodis: Now using nulled macro instead for asserts. 
Moved the nulled macro to global.h for the time being.

notyourav (approved):

Pheenoh (approved):

PR #62, opened 2021-01-04
some daAlink_c funcs

TakaRikka:
checkMagicArmorHeavy, checkBootsOrArmorHeavy, setSpecialGravity, onFrollCrashFlg, setMidnaTalkStatus ok


notyourav (approved):
Chock full of inlines but those can be revisited later

include/d/d_a/d_a_alink/d_a_alink.h
```diff
@@ -44,6 +44,12 @@ struct csXyz {  // move later
 
 class daAlink_c {
 public:
+    // inlines
+    void dComIfGp_setZStatus(u8 status, u8 unk) {
```
Pheenoh, line 48: Because of their naming convention, these inlines are probably actually defined in d_com_inf_game.h. We can leave it for now but as we start going back and defining these we should probably do it there

PR #65, opened 2021-01-05
add a lot of dComIfGs inlines

TakaRikka:


include/d/d_com/d_com_inf_game/d_com_inf_game.h
```diff
@@ -166,6 +181,151 @@ class dComIfG_inf_c {
...
+inline void dComIfGs_setRodTypeLevelUp(void) {
+    g_dComIfG_gameInfo.getSaveFile().getPlayerItem().setRodTypeLevelUp();
+}
+inline void dComIfGs_setArrowNum(u8 amount) {
+    g_dComIfG_gameInfo.getSaveFile().getPlayerItemRecord().setArrowNum(amount);
+}
+inline void dComIfGs_setArrowMax(u8 max) {
+    g_dComIfG_gameInfo.getSaveFile().getPlayerItemMax().setArrowNum(max);
+}
+inline u8 dComIfGs_getPachinkoMax(void) {
```
notyourav, line 276: Have you tried compiling these without the explicit `inline`? Unless its necessary, it's cleaner without them.
```suggestion 
u8 dComIfGs_getPachinkoMax(void) {
```

notyourav (changes requested):

include/d/d_com/d_com_inf_game/d_com_inf_game.h
TakaRikka: yea the way they are like this, compiling without inline makes it not match

notyourav (approved):

PR #69, opened 2021-01-05
JKRSolidHeap

Julgodis:
* decompiled JKRSolidHeap (a few non-matching functions left)
* added const to virtual functions
* JKRHeap::TState


include/JSystem/JKernel/JKRAssertHeap/JKRAssertHeap.h
```diff
@@ -10,23 +10,23 @@ class JKRAssertHeap : public JKRHeap {
...
-    /* vt[18] */ virtual s32 do_getTotalFreeSize(void);   /* override */
-    /* vt[19] */ virtual u8 do_changeGroupID(u8 param_1); /* override */
-    /* vt[20] */ virtual u8 do_getCurrentGroupId(void);   /* override */
+    /* vt[04] */ virtual u32 getHeapType(void) const;          /* override */
+    /* vt[05] */ virtual bool check(void);                     /* override */
+    /* vt[06] */ virtual bool dump_sort(void);                 /* override */
+    /* vt[07] */ virtual bool dump(void);                      /* override */
+    /* vt[08] */ virtual void do_destroy(void);                /* override */
+    /* vt[09] */ virtual void* do_alloc(u32, int);             /* override */
+    /* vt[10] */ virtual void do_free(void*);                  /* override */
```
notyourav, line 19:
```suggestion
    /* vt[10] */ void do_free(void*);                  /* override */
```
Do these need to be marked as virtual?

libs/JSystem/JKernel/JKRSolidHeap.cpp
```diff
@@ -1,112 +1,302 @@
...
+        heap = getRootHeap();
+    }
+
+    if (size == 0xffffffff) {
+        size = heap->getMaxAllocatableSize(0x10);
+    }
+
+    u32 alignedSize = ALIGN_PREV(size, 0x10);
+    u32 solidHeapSize = ALIGN_NEXT(sizeof(JKRSolidHeap), 0x10);
+    if (alignedSize < solidHeapSize) {
```
notyourav, line 15:
```suggestion
    if (alignedSize < solidHeapSize)
```
for consistency, or you should add parentheses around the condition below.

notyourav (changes requested):

include/JSystem/JKernel/JKRAssertHeap/JKRAssertHeap.h
Julgodis: For function that are overridden, virtual is not required. With both the vt[XX] and override comment maybe the virtual keyword is not needed, but this is how I done it for everything else in JKernel. Best would be to use the override keyword, but that was release in C++11, which MWCC ofcourse doesn't support. I am fine with doing it either way as this doesn't really effect the matching of functions. But maybe it's better to change it later, in all classes at once, when we have decided on the style we want.

notyourav (approved):

PR #72, opened 2021-01-06
match some Z2Calc functions

lepelog:
`Z2Calc::getRandom_0_1` is pretty much just `get_ufloat_1` inlined, which doesn't work because it adds an additional 1.0f to data. Added a note so it can be fixed in the future


include/JSystem/JMath/random.h
```diff
@@ -8,6 +8,23 @@ struct TRandom_fast_ {
...
 
     TRandom_fast_(u32 value);
+    u32 get(void) {
+        value = (value * 0x19660d) + 0x3c6ef35f;
+        return value;
+    }
+
+    u32 get_bit32(void) { return this->get(); }
+
+    float get_ufloat_1(void) {
```
barzamin, line 18: did you forget to check in the mentioned note?

include/JSystem/JMath/random.h
```diff
@@ -8,6 +8,23 @@ struct TRandom_fast_ {
...
+        union {
+            f32 f;
+            s32 s;
+        } out;
+        out.s = ((this->get() >> 9) & 0x7FFFFF) | 0x3f800000;
```
leoetlino, line 23: Just wondering, is the `& 0x7FFFFF` actually needed? And does this still match with u32 instead of s32?

On a related note, this kind of type punning is undefined behavior in standard C++.

barzamin: i wouldn't be too concerned about it being UB in standard C atm since this is just a way of memeing something MWCC would generate internally

lepelog: u32 works as well, is there a reason to use one over the other? `& 0x7FFFFF` isn't needed. Is there a better way to write it that is not UB?

leoetlino: u32 would be preferable since this is doing bit manipulation, and get() returns a u32 so you avoid one conversion.

The fully standards-compliant method would be to do:

```cpp
float value;
std::memcpy(&value, u32_value, sizeof(value));
```

But older compilers like mwcc might be incapable of optimising out the memcpy call... If this doesn't work, I would suggest adding a "this is UB" comment (similar to "this is a bug" comments)

Pheenoh (approved):

PR #78, opened 2021-01-08
JKernel: static class variables and more matches

Julgodis:
* abs, memset
* changed JKRCompression to typedef (this will make signatures make)
* changed some static variables in JKernel from lbl_XXXXXXXX to their real name
* JKRArchive setExpandSize and getExpandSize matched
* JKRDvdFile 100% complete
* JKRDvdArchive matched a few, still problems with instruction ordering


include/JSystem/JKernel/SArc.h
```diff
@@ -0,0 +1,64 @@
...
+struct SArcHeader {
+    u32 signature;
+    u32 file_length;
+    u32 header_length;
+    u32 file_data_offset;
+    u32 file_data_length;
+    u32 field_0x14;
+    u32 field_0x18;
+    u32 field_0x1c;
+} __attribute__((aligned(32)));
```
notyourav, line 15: Is `__attribute__` valid in mwcc? Curious if removing this changes anything, or if its simply being ignored by the preprocessor.

libs/JSystem/JKernel/JKRHeap.cpp
```diff
@@ -80,13 +79,13 @@ bool JKRHeap::initArena(char** memory, u32* size, int param_3) {
...
-    ram_start = (ram + 0x1fU & 0xffffffe0);
-    ram_end = (high & 0xffffffe0);
-    lbl_80451384 = OS_GLOBAL_ADDR(void, 0x80000000);
-    lbl_80451388 = (void*)ram_start;
-    lbl_8045138C = (void*)ram_start;
-    lbl_80451390 = (void*)ram_end;
-    lbl_80451394 = OS_GLOBAL(u32, 0x80000028);
+    ram_start = ALIGN_NEXT(ram, 0x20);
+    ram_end = ALIGN_PREV(high, 0x20);
+    mCodeStart = OS_GLOBAL_ADDR(void, 0x80000000);
```
notyourav, line 84:
```suggestion
    mCodeStart = OSPhysicalToCached(void);
```
inline

libs/JSystem/JKernel/JKRHeap.cpp
```diff
@@ -80,13 +79,13 @@ bool JKRHeap::initArena(char** memory, u32* size, int param_3) {
...
+    mCodeEnd = (void*)ram_start;
+    mUserRamStart = (void*)ram_start;
+    mUserRamEnd = (void*)ram_end;
+    mMemorySize = OS_GLOBAL(u32, 0x80000028);
```
notyourav, line 88: I think this might be a struct offset from the address retrieved above (`0x80000000`)
Ghidra outputs (debug):
```cpp
_mMemorySize = *(undefined4 *)(_mCodeStart + 0x28);
```
Also the inline for retrieving the right address is only called once.

include/global.h
```diff
@@ -11,6 +11,7 @@
 #define JUT_ASSERT(...)
 #define ASSERT(...)
 #define LOGF(FMT, ...)
+#define FLAG_HAS(V, F) (((V) & (F)) == 0)
```
notyourav, line 14:
```suggestion
#define FLAG_ON(V, F) (((V) & (F)) == 0)
```
might be more verbose

notyourav (changes requested):

include/JSystem/JKernel/SArc.h
Julgodis: Don't known if it's valid or not, tried it in the function JKRDvdArchive::fetchResource_subroutine. Forgot to remove it before committing, good catch!

notyourav (approved):

Julgodis:
These changed are already merged. As pull request #80 was based on this pull request, and #80 is already merged.

PR #94, opened 2021-01-24
decomp DynamicLink

barzamin:
wip


src/DynamicLink.cpp
```diff
@@ -0,0 +1,334 @@
...
+        if (do_link() == 0) {
+            return false;
+        }
+
+        if (field_0x2 < 0xffff) {
+            field_0x2++;
+        }
+    }
+
+    JUT_ASSERT(mLinkCount != 0xffff);  // "mLinkCount < 65535"
```
notyourav, line 72: If they originally wrote this as 65535 then we should do the same

barzamin:
oh this is going to be so out-of-date

PR #98, opened 2021-01-25
change split.py to output extern vars/functions into generate CPP

lepelog:
also change demangling to use short type names


tools/splitter/split.py
```diff
@@ -331,10 +326,6 @@ def find_labels_in_operands(operands):
             logger.info(f'adding extern func {label} to {extern_functions_file}')
             funcs_new_labels.add(label)
 
-    if len(funcs_new_labels) > 0:
-        with open(extern_functions_file, 'a') as f:
-            f.write('\n\n')
-            f.write(emit_cxx_extern_fns(cxx_out.name, funcs_new_labels))
```
barzamin, line 337: remove corresponding `@click.option` and argument to `split()`

tools/splitter/split.py
```diff
@@ -270,11 +270,6 @@ def find_labels_in_operands(operands):
             logger.debug(f'adding extern var {label} to {extern_variables_file}')
             vars_new.add(label)
 
-    if len(vars_new) > 0:
-        with open(extern_variables_file, 'a') as f:
-            f.write('\n\n')
-            f.write(emit_cxx_extern_vars(cxx_out.name, vars_new))
-
```
barzamin, line 277: remove corresponding `@click.option` and argument to `split()`

barzamin (changes requested):

Pheenoh (approved):

PR #107, opened 2021-01-29
Tool for extracting literals and strings

Julgodis:
* New tool `section2cpp.py` for extracting literals, strings, and other data from baserom.dol
* Replaced a few string labels with "real" strings

rodata:
```bash
> ./tools/section2cpp.py --section .rodata --string --object JKRSolidHeap.o
```
```c++
// JKRSolidHeap.o
static const char lbl_8039CE50 = "allocFromHead: cannot alloc memory (0x%x byte).\n";
static const char lbl_8039CE81 = "allocFromTail: cannot alloc memory (0x%x byte).\n";
static const char lbl_8039CEB2 = "free: cannot free memory block (%08x)\n";
static const char lbl_8039CED9 = "resize: cannot resize memory block (%08x: %d)\n";
static const char lbl_8039CF08 = "getSize: cannot get memory block size (%08x)\n";
static const char lbl_8039CF36 = "check: bad total memory block size (%08X, %08X)\n";
static const char lbl_8039CF67 = "head %08x: %08x\n";
static const char lbl_8039CF78 = "tail %08x: %08x\n";
static const char lbl_8039CF89 = "%d / %d bytes (%6.2f%%) used\n";
static const char lbl_8039CFA7 = "";                                                       // padding
```

sdata2:
```bash
> ./tools/section2cpp.py --section .sdata2 --object JKRSolidHeap.o
```
```c++
// JKRSolidHeap.o
static const float lbl_80455FA8 = 100.0f;                                                  // 0x42c80000
static const u8 lbl_80455FAC[4] = { 0x00, 0x00, 0x00, 0x00 };                              // padding
static const double lbl_80455FB0 = 4503601774854144.0;                                     // 0x4330000080000000 | s32 to float (compiler-generated)
static const double lbl_80455FB8 = 4503599627370496.0;                                     // 0x4330000000000000 | u32 to float (compiler-generated)
```

### DIFF
--- a/src/m_Do/m_Do_main.cpp
+++ b/src/m_Do/m_Do_main.cpp
@@ -22,12 +22,13 @@ extern u8 lbl_80451A18;
 extern char lbl_803739A0;
 
 void version_check(void) {
-    if ((!strcmp("20Apr2004", "20Apr2004")) &&
-        (!strcmp("Patch2", "Patch2"))) {
+    if ((!strcmp("20Apr2004", "20Apr2004")) && (!strcmp("Patch2", "Patch2"))) {
         return;
     }
 
-    OSReport_Error((char*)"\x53\x44\x4B\x82\xCC\x83\x6F\x81\x5B\x83\x57\x83\x87\x83\x93\x82\xAA\x88\xEA\x92\x76\x82\xB5\x82\xDC\x82\xB9\x82\xF1\x81\x42\x92\xE2\x8E\x7E\x82\xB5\x82\xDC\x82\xB7\x0A");
+    OSReport_Error((char*)"\x53\x44\x4B\x82\xCC\x83\x6F\x81\x5B\x83\x57\x83\x87\x83\x93\x82\xAA\x88"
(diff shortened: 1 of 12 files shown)


src/m_Do/m_Do_main.cpp
```diff
@@ -122,30 +155,42 @@ asm void Debug_console(u32) {
...
 #else
+const char* lbl_80373C23 = "/str/Final/Release/COPYDATE";
 asm void LOAD_COPYDATE(void*) {
     nofralloc
 #include "m_Do\m_Do_main\asm\func_8000614C.s"
 }
 #endif
 
+const char* lbl_80373C3F = "Root";
+const char* lbl_80373C44 = "\x83\x8B\x81\x5B\x83\x67"; /* undecodable string */
```
notyourav, line 178: These are all SJIS strings.

Julgodis: Yeah I know, the tool can decode them (undecodable string is wrong).
Have not figured out how to get them working with Metrowerks yet, and this was my solution for now.

notyourav (approved):

PR #112, opened 2021-02-08
f_op_actor_mng split & started decomp

lepelog:


include/JSystem/JMath/JMath.h
```diff
@@ -0,0 +1,21 @@
...
+
+struct sinCosTableEntry {
+    f32 sin;
+    f32 cos;
+};
+
+extern sinCosTableEntry lbl_80439A20[0x2000];  // JMath::JMath::sincosTable_
+
+inline f32 cM_scos(s16 s) {
+    return lbl_80439A20[(s & 0xFFFF) >> 3].cos;
```
notyourav, line 14:
```suggestion
    return lbl_80439A20[s >> 3].cos;
```
is this needed?

lepelog: That change doesn't match, however `return lbl_80439A20[static_cast<u16>(s) >> 3].cos;` also matches, would that be preferred?

notyourav: Hmm.. I would say yes, given the context (no negative array access). They may have had one here.

notyourav (approved):

include/JSystem/JMath/JMath.h
lepelog: Looking at debug, they had way more layers of inlined (template) functions here

PR #117, opened 2021-02-27
Match a few nonmatchings and delete unused asm

lepelog:


src/f/f_op/f_op_scene_iter.cpp
```diff
@@ -1,21 +1,21 @@
...
+extern node_list_class lbl_803A3918;  // f_op_scene_tag::g_fopScnTg_SceneList
 
-asm void* fopScnIt_Judge(fop_ScnItFunc pFunc1, void* pFunc2) {
-    nofralloc
-#include "f/f_op/f_op_scene_iter/asm/func_8001EC74.s"
+void* fopScnIt_Judge(fop_ScnItFunc pFunc1, void* pData) {
+    struct {
+        fop_ScnItFunc mFunc;
+        void* mpData;
+    } iterParams;
```
notyourav, line 17: Is there anywhere else this is used? A bit unusual to have an inline definition like this

lepelog: I don't think so? It also doesn't matter for matching if it's an anonymous struct like this or a named one outside of the function. This is how we have it currently for all the other iter functions like this

notyourav: Gotcha, I wasn't aware of that. Perhaps that can be addressed in a different commit.

notyourav (approved):

src/f/f_op/f_op_scene_iter.cpp
lepelog: Seems like I'm misremembering it, in f_op_actor_iter it's written like this, but the iters in f_pc have that struct named and outside of the function

PR #118, opened 2021-03-05
Fix symbol names

lepelog:
used script: https://gist.github.com/lepelog/9d560148e4caf57cf81ab4e1820dbc79

Renames most function symbols to match the original map file.  
Still missing:
- Some JKRHeap functions, but the changes look intentional
- Some fpc functions, mostly long vs int and function pointer params, will probably fix those in a separate PR
- `__register_global_object` and `__destroy_global_chain` cause the linker complained about them
- All functions with mangled symbols that include `<` or `@`


notyourav:
> changes look intentional

Could you elaborate on this? My guess is those symbols were done before we were too concerned about accuracy to the original map file.

notyourav:
> mangled symbols that include < or @

Is there any Hope here? I'm guessing anything that has references within inline asm won't be happy. Which could be all eventually

Pheenoh (approved):

PR #120, opened 2021-03-28
horse_static match / itembase_static partial match

TakaRikka:
+ some small clean up


include/d/d_a/d_a_horse_static/d_a_horse_static.h
```diff
@@ -1,19 +1,21 @@
-#include "dolphin/types.h"
+#include "Z2AudioLib/Z2Creature/Z2Creature.h"
```
jdflyer, line 1: Make sure you add include guards here

Pheenoh (approved):

PR #121, opened 2021-03-28
Full split of main.dol and RELs

Julgodis:
This pull-request uses the new dol2asm tool to split the main.dol and the .rel files into C++ and asm code.

### Scripts:
**dol2asm.py:** This tool splits executable files to C++ and asm code. The tool will use the linker maps and the executable code to find symbols and determine their types. For each translation unit, the tool will output the following: types, forward references, external references, function and data. Most function names are demangled (if they don’t use some weird characters). References in and between to other symbols are found, including between RELs.

The tool will generate makefiles (per library and REL) and, what I am calling, a symbol definition file for each module.  The symbol definition file is a python file with information about all the symbol the tool could find and generated. These files are used to generate the ldscript.lcf file. The code generated by the tool will also fix some of the previous issue. E.g., it uses a different instruction (or a simplified mnemonic) to get around the SDA hack. Another big problem that this tool solves is that the sections are now the correct size. The .dol format aligns everything to 32 bytes, but the actual sections are not aligned. Lucky this information is available because the linker writes the section address and size for each section into `_rom_copy_info` and `_bss_init_info`.

Because the tool is using more then the main.dol (baserom.dol) all the assets are now required to be located in the folder `game`.

Running the dol2asm tool is simple (although, it is not very fast and takes about 10 minutes and uses a lot of ram). It uses a few python dependencies (such as numpy, rich, etc).
```bash
python3 tools/dol2asm.py
```

**lcf.py:** This script will use the symbol definition files to generate a ldscript.lcf file. By generating the ldscript file at link-time we remove the need to manual use FORCEACTIVE. Additionally, some cool ticks are used to get strings and function with bad names working.

**tp.py:** This helper script is used to calculate the current decompiled progress and to help make sure everything works before submitting a pull-request.

**makerel.py:** Script will take all .plf files and the static main.elf binary and generate .rel files.

**rel.py:** Script functions like the objdump/readelf programs and print information about .rel files. 

image

All current decompiled code need to merged before mering this with master. Meanwhile, this should be merged to a new barnch where everyone can contribute to port the code over. Before that these things should probably be figured out.
- [X] RELs decompiled and building.
- [X] Decompiled progress for RELs.
- [x]  Test and make sure `diff.py` works

Example of generated code:
image

#### Changes:
* Decompile data as floats and doubles
* Fixed .dead striping making everything FORCE-ACTIVE
* Added support for decompiling __sinit functions in any order
* Updated progress script
* Building and generating matching RELs
* Auto decompiling small functions
* Split stringBase0 to make it easier to decompile one function at the time


Pheenoh:
Added this as a new branch. Let's work off that until we have all the code ported over.

PR #122, opened 2021-04-02
fix dComIfG_play_c

TakaRikka:
dComIfG_play_c should be mostly accurate with this. some types or class sizes may be wrong in some places, but should be much closer than before

### DIFF
--- a/include/d/d_bg/d_bg_s/d_bg_s.h
+++ b/include/d/d_bg/d_bg_s/d_bg_s.h
@@ -39,7 +39,6 @@ public:
 private:
     /* 0x0000 */ cBgS cbgs;
     /* 0x1400 */ void* vtable;
-    // 0x1408 dBgS_HIO dbgs_hio; only on debug rom?
 };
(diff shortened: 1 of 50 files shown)


include/d/d_bg/d_bg_s/d_bg_s.h
```diff
@@ -38,8 +38,8 @@ class dBgS {
 public:
 private:
     /* 0x0000 */ cBgS cbgs;
-    /* 0x1400 */ u8 vtable[8];
-    /* 0x1408 */ dBgS_HIO dbgs_hio;
+    /* 0x1400 */ void* vtable;
+    // 0x1408 dBgS_HIO dbgs_hio; only on debug rom?
 };
 #pragma pack(pop)
```
notyourav, line 44: Is this still required?

TakaRikka: nah forgot that

notyourav (approved):

PR #124, opened 2021-04-26
fix format check by not checking tools folder for clang-format

lepelog:


lepelog:
Just realized the clang-format check is already fixed with https://github.com/zeldaret/tp/pull/123/commits/192f87e1d66f13bf309cb9fc6972c2150bcacaa8

PR #130, opened 2021-05-04
move d_a_alink and some related stuff

TakaRikka:


src/m_Do/m_Do_ext.cpp
```diff
@@ -80,7 +74,7 @@ struct Z2Creature {
     /* 802C0628 */ void initAnime(void*, bool, f32, f32);
     /* 802C06D0 */ void updateAnime(f32, f32);
 };
-
+//
```
lepelog, line 77: I guess this can be removed

lepelog (approved):

PR #134, opened 2021-06-13
add missing config options

kipcode66:
This PR adds corrections and renaming of options which are used only for Wii saves. This includes fixing the the item buttons, and the specification of some configuration options that were flagged as "unknown".

If there's an issue with the modifications' format, I'm open to criticism.

### DIFF
--- a/include/d/d_save/d_save/d_save.h
+++ b/include/d/d_save/d_save/d_save.h
@@ -7,10 +7,10 @@
 #include "global.h"
 
 #define DEFAULT_SELECT_ITEM_INDEX 0
-#define MAX_SELECT_ITEM 3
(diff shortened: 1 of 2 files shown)


TakaRikka (commented):
this should probably be pr'd towards the full split branch, since the master branch is outdated now. the member var names look good, though the functions probably shouldnt be added for building GC, or at least made inline

kipcode66:
When I compiled it, since the code was not used, it didn't change the final DOL

Pheenoh:
The fullsplit branch will be used as the primary branch going forward until all the previously decomped code is moved over from master. Please migrate these changes to that branch and reopen.

PR #135, opened 2021-06-14
fix item slots in config

kipcode66:
This PR fixes the item slots for the buttons and change the names of some fields for the options for names that are closer to the one used in the game's UI.


TakaRikka:
personally prefer keeping the old variable names because it matches what the inlined functions call them in debug rom, but there isn't a hard rule set so maybe others can give input

kipcode66:
I feel the names for the calibration pointers are more meaningful, but for the rest of them it's not necessary and I can remove those change. The main thing was the item slot fix.

magcius:
Are you sure that your names for the calibration values are correct? I wouldn't normally expect X/Y to have different types, but one is u16, and one is u8.

kipcode66:
@magcius When I was calibrating the pointer while looking at the data, one was controlling the horizontal calibration, the other was for the vertical. And yes, The vertical value has a lot less resolution than the horizontal one.

magcius:
I would prefer to keep the game's names for now

kipcode66:
where did these name came from? I don't see them in the map file.

TakaRikka:
they're from the debug rom. such as dSv_player_config_c::setCalibrateDist. there's a lot of these types of functions that were inlined in retail

kipcode66:
All the new code that was merged recently made a mess in this PR. So I will close this PR and open an other one with the item slot fix only.

PR #138, opened 2021-07-27
some c_cc and d_cc stuff

lepelog:
`d_cc_d` doesn't match because of reorderings of inline vtable functions. Otherwise the file is decompiled, so the nonmatching version in now called `d_cc_d_nonmatching.cpp`


lepelog:
Included in #140

PR #141, opened 2021-08-18
move more of d_save / d_com_inf_game

TakaRikka:
cleaned up a bunch of stuff + some new matches too


lepelog (changes requested):
Looks good, just a few things I noticed. Also, some functions still use the mangled names (like https://github.com/zeldaret/tp/blob/1a75f14acb14be13cd5a8c727a00964acee89227/src/d/d_item.cpp#L1191 ), I think it would be good to change them to the demangled names if possible

lepelog (approved):

src/d/com/d_com_inf_game.cpp
```diff
@@ -1994,26 +1996,71 @@ u8 dComIfGs_getBottleMax() {
...
-asm void dComIfGp_getSelectItemMaxNum(int param_0) {
-    nofralloc
-#include "asm/d/com/d_com_inf_game/dComIfGp_getSelectItemMaxNum__Fi.s"
+int dComIfGp_getSelectItemMaxNum(int param_0) {
+    u8 selectItem = dComIfGp_getSelectItem(param_0);
+    int itemNum = 0;
+
+    if (selectItem == BOMB_BAG_LV1) {
+        itemNum = 1;
+    } else if ((u8)(selectItem - NORMAL_BOMB) <= 2 || selectItem == BOMB_ARROW) {
```
lepelog, line 2024: `(u8)(selectItem - NORMAL_BOMB) <= 2` is a compiler optimization for checking if selectItem is equal to `NORMAL_BOMB` or the 2 values after it in the enum. It should match with 3 separate comparisons as well and that's more readable imo

ProjectRevoTPP: I think I already posted about this in the discord a while back as it's needed for isBomb.

TakaRikka: yea i prob missed it before. good to know now

PR #145, opened 2021-09-04
m_Do cleanup, d_resource, drawlist stuff

TakaRikka:
/ some work setting up various classes needed for other stuff


src/d/event/d_event.cpp
```diff
@@ -682,38 +715,35 @@ dEvt_info_c::dEvt_info_c() {
...
 
 /* 80043480-800434CC 03DDC0 004C+00 1/1 0/0 0/0 .text            getEventName__11dEvt_info_cFv */
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-asm void dEvt_info_c::getEventName() {
-    nofralloc
-#include "asm/d/event/d_event/getEventName__11dEvt_info_cFv.s"
+// fake match?
+s16 dEvt_info_c::getEventName() {
```
lepelog, line 728: I don't think it returns a s16. Is the first member of `dEvDtEvent_c` a string? Cause then it probably returns a pointer to that

src/m_Do/m_Do_lib.cpp
```diff
@@ -92,6 +98,21 @@ f32 mDoLib_clipper::mFovyRate;
...
+    mClipper.setFovy(fovy);
+    mClipper.setAspect(aspect);
+    mClipper.setNear(near);
+    mClipper.setFar(far);
+    mSystemFar = far;
+    mClipper.calcViewFrustum();
+
+    int tmp = param_0 * 182.04444885253906f;
+
+    mFovyRate = (JMath::sincosTable_.table[tmp].b1) / (JMath::sincosTable_.table[tmp].a1);
```
lepelog, line 113: Pretty sure these are `cM_ssin` and `cM_scos`

libs/JSystem/JUtility/JUTXfb.cpp
```diff
@@ -102,59 +81,61 @@ asm JUTXfb::JUTXfb(_GXRenderModeObj const* param_0, JKRHeap* param_1, JUTXfb::EX
...
-#pragma optimizewithasm off
-asm void JUTXfb::initiate(u16 param_0, u16 param_1, JKRHeap* param_2, JUTXfb::EXfbNumber param_3) {
-    nofralloc
-#include "asm/JSystem/JUtility/JUTXfb/initiate__6JUTXfbFUsUsP7JKRHeapQ26JUTXfb10EXfbNumber.s"
+void JUTXfb::initiate(u16 width, u16 height, JKRHeap* pHeap, JUTXfb::EXfbNumber xfbNum) {
+    if (pHeap == NULL) {
+        pHeap = JKRHeap::getSystemHeap();
+    }
+
+    int size = ((u32)width + 0xf & 0xfff0) * (u32)height * 2;
```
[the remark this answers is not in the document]
TakaRikka: this does a `rlwinm r3, r0, 0, 0, 0x1b` instead of `rlwinm r3, r0, 0, 0x10, 0x1b`

src/d/event/d_event.cpp
TakaRikka: this could be the case, but adding a data return adds extra load instructions. not sure if theres some way to get rid of them.

Pheenoh (approved):

PR #149, opened 2021-09-26
JUTDirectPrint

Julgodis:
Added support for: va_list, va_start, va_end.
Removed unused .s files.

Decompiled:
* __ct__14JUTDirectPrintFv
* start__14JUTDirectPrintFv
* erase__14JUTDirectPrintFiiii
* changeFrameBuffer__14JUTDirectPrintFPvUsUs
* printSub__14JUTDirectPrintFUsUsPCcP16__va_list_structb
* drawString__14JUTDirectPrintFUsUsPc
* drawString_f__14JUTDirectPrintFUsUsPCce
* setCharColor__14JUTDirectPrintFQ28JUtility6TColor
* setCharColor__14JUTDirectPrintFUcUcUc

Decompiled (nonmatching):
* drawChar__14JUTDirectPrintFiii


include/JSystem/JUtility/JUTDirectPrint.h
```diff
@@ -1,6 +1,59 @@
...
+    /* 802E456C */ void changeFrameBuffer(void*, u16, u16);
+    /* 802E45A4 */ void printSub(u16, u16, char const*, __va_list_struct*, bool);
+    /* 802E46D8 */ void drawString(u16, u16, char*);
+    /* 802E4708 */ void drawString_f(u16, u16, char const*, ...);
+    /* 802E47C8 */ void setCharColor(u8, u8, u8);
+    /* 802E4798 */ void setCharColor(JUtility::TColor);
+
+    /* 802E4240 */ static JUTDirectPrint* start();
+
+#if DEBUG
```
TakaRikka, line 27: i don't think we've had any specific rule on handling debug stuff, but i've been just including necessary debug stuff as inlines and not making any distinction in the code. not sure what future plans there are for multiple version support

include/JSystem/JUtility/TColor.h
```diff
@@ -0,0 +1,44 @@
+#ifndef TCOLOR_H
+#define TCOLOR_H
+
+#include "dolphin/gx/GXTexture.h"
+#include "dolphin/types.h"
+
+namespace JUtility {
```
TakaRikka, line 7: this is duplicate info from "include/JSystem/JUtility.h". maybe should merge them?

libs/JSystem/JUtility/JUTDirectPrint.cpp
```diff
@@ -321,113 +257,104 @@ asm void JUTDirectPrint::drawChar(int param_0, int param_1, int param_2) {
...
 
-/* ############################################################################################## */
-/* 8039D9A0-8039D9A0 02A000 0000+00 0/0 0/0 0/0 .rodata          @stringBase0 */
-#pragma push
-#pragma force_active on
-SECTION_DEAD static char const* const stringBase_8039D9A0 = "%s";
-/* @stringBase0 padding */
-SECTION_DEAD static char const* const pad_8039D9A3 = "\0\0\0\0";
-#pragma pop
+#if DEBUG
```
TakaRikka, line 311: same comment about debug related things

include/JSystem/JUtility/TColor.h
Julgodis: Fixed. Good catch!

include/JSystem/JUtility/JUTDirectPrint.h
Julgodis: In this case it didn't matter, OK with or without `#if DEBUG` (see bb9dbd7). Maybe we should create some simple rules for handling the debug stuff? For most debug functions, adding inline to them will probably work, but for ASSERT and other macros this may not be the case.

TakaRikka: oh i see, yeah probably should make some rules about how to handle that then for where it is needed

TakaRikka (approved):

PR #162, opened 2021-11-28
reduce python dependencies to just build

lepelog:
numpy and capstone are only required for disassembling and splitting the dol, so don't install and require them by default

### DIFF
--- a/tools/requirements.txt
+++ b/tools/requirements.txt
@@ -1,8 +1,4 @@
 rich
 click
-intervaltree
 yaz0
-numpy
-capstone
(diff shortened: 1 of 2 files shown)


tools/requirements.txt
```diff
@@ -1,8 +1,4 @@
 rich
 click
-intervaltree
 yaz0
-numpy
-capstone
-aiofiles
-GitPython
\ No newline at end of file
+GitPython
```
jdflyer, line 4: Should we also add the required dependencies for diff.py here?

PR #163, opened 2021-11-29
Clean up and improvements to tools

Julgodis:
This PR does some needed cleanup of the tools and the python code. There are a few tools, which I don't believe are being used, that has now been removed (if you were using something that got removed let me know and I will revert the change). E.g. `postprocess.py` hasn't been used since the move to the dol2asm code. 

To improve the setup process, when starting from a clean repository, there is now a new command for the `./tp` program. The `./tp setup` command will setup the project by doing the following:
- create directory structure
- make sure you have the compiler downloaded
- patch the compiler if not already done (no need to download the patched compiler anymore)
- extract game assets using the `extract_game_assets.py` script

There are performance improvements to both `lcf.py` (which no longer need to import the whole dol2asm package) and `./tp progress` (where you now can get the progress for the whole project, include RELs, without waiting for 10s of seconds). 

For a while now `./tp` has a few nice commands which I believe should be use more often. E.g.,  `./tp pull-request` will format all code, build the project (including the RELs), delete unused `.s` files, and check that everything is OK. There always seems to be `.s` files that are not deleted which should never be a problem. If everyone were to use the `pull-request` command this problem would go way, otherwise maybe do some Github action to check for "undeleted" `.s` files. 

_The README will need to be updated after this is merged_

Removed:
* `postprocess.py`
* `vtables.py`
* `find_unused_asm.py`
* `section2cpp.py`
* `splitter/`
* `difftools.py`

Changes/New:
* `./tp setup` 
* `requirements.txt` - Update to include all necessary packages for diff.py etc.
* `conflict.py` - Finds conflicts which causes the build to not be OK, e.g., symbol moved to different section.
* `Makefile` - Fixed the problem where `__sinit` functions would have the incorrect name, e.g, `__sinit_m_Do_lib_o_iconv_cpp` instead of `__sinit_m_Do_lib_cpp`.

_This PR also does the same thing as #162_


Julgodis:
Because `OK Check` action isn't configured to use the new path of the patched compiler the build does fail. 
The patched compiler is now expected to be at `tools\mwcc_compiler\2.7\mwcceppc_patched.exe`

lepelog:
Nice job with this, should make it a lot easier for people to get started as well.

My only issue with `./tp pull-request` is, that it builds all of the rels as well, even though they aren't being decompiled right now. I think it would be a good idea to have an option to skip building the rels, at least until we actually start decompiling them. Might be only a problem for me, because my laptop is so slow.

I think it's a good idea to check for not deleted asm files in github actions as well, maybe with adding a `--check` option to the `remove-unused-asm" command?

TakaRikka:
it might be nice if there was a separate `clean` option that only cleans up the main dol, since building all rels takes quite a while

Julgodis:
> it might be nice if there was a separate `clean` option that only cleans up the main dol, since building all rels takes quite a while

I changed the default `make clean` to only remove files related to building `main.dol`. Added two new targets, `make clean_rels` will remove all associated build artifacts for RELs and `make clean_all` will remove  build artifacts for both `main.dol` and RELs.

> Nice job with this, should make it a lot easier for people to get started as well.
> 
> My only issue with `./tp pull-request` is, that it builds all of the rels as well, even though they aren't being decompiled right now. I think it would be a good idea to have an option to skip building the rels, at least until we actually start decompiling them. Might be only a problem for me, because my laptop is so slow.
> 
> I think it's a good idea to check for not deleted asm files in github actions as well, maybe with adding a `--check` option to the `remove-unused-asm" command?

Now RELs are not included by the default in `./tp pull-request` and `./tp check`. To included them again use the `--rels` option. `./tp remove-unused-asm --check` will now check if there exists unused asm files that can be remove. If there exists files the program will exit with 1, otherwise it will print "OK" and exit with 0.

TakaRikka (approved):

lepelog (approved):

PR #164, opened 2021-12-03
JUTDbPrint

lepelog:


libs/JSystem/JUtility/JUTDbPrint.cpp
```diff
@@ -6,167 +6,135 @@
 #include "JSystem/JUtility/JUTDbPrint.h"
 #include "dol2asm.h"
 #include "dolphin/types.h"
+#include "MSL_C.PPCEABI.bare.H/MSL_Common/Src/string.h"
```
notyourav, line 9: do we have a better header to use?

libs/JSystem/JUtility/JUTDbPrint.cpp
```diff
@@ -6,167 +6,135 @@
...
-struct J2DOrthoGraph {
-    /* 802E96D0 */ J2DOrthoGraph(f32, f32, f32, f32, f32, f32);
-    /* 802E97B4 */ void setPort();
+class J2DOrthoGraph : public J2DGrafContext {
+public:
+    char fake[0xd0];
+    J2DOrthoGraph(f32, f32, f32, f32, f32, f32);
+    virtual ~J2DOrthoGraph() {};
+    void setPort();
 };
```
TakaRikka, line 24: are there issues with the classes in the J2D headers or can these be removed for those instead?

lepelog: Changing `virtual ~J2DGrafContext() {};` caused problems I wasn't able to resolve but I didn't try too hard

include/JSystem/JUtility/JUTVideo.h
```diff
@@ -25,6 +25,9 @@ class JUTVideo {
     /* 802E5144 */ static void postRetraceProc(u32);
     /* 802E50BC */ static void drawDoneCallback();
 
+    u32 getFbWidth() const { return this->mRenderObj->fb_width; }
+    u32 getEfbHeight() const { return this->mRenderObj->efb_height; }
```
Julgodis, line 29: remove `this->`
```c++
u32 getFbWidth() const { return mRenderObj->fb_width; }
u32 getEfbHeight() const { return mRenderObj->efb_height; }
```

TakaRikka (approved):

PR #171, opened 2021-12-27
d_a_player / d_meter_map + misc

TakaRikka:
most of d_a_player decompiled now. d_meter_map mostly matching, but float data + few small issues leaving many as nonmatching for now. misc d_meter / d_a_alink work


src/d/meter/d_meter_map.cpp
```diff
@@ -132,93 +78,178 @@ extern "C" void _savegpr_29();
...
+inline BOOL dComIfGp_event_runCheck() {
+    return g_dComIfG_gameInfo.play.getEvent().runCheck();
+}
+
+inline u16 dComIfGp_event_checkHind(u16 flag) {
+    if (!dComIfGp_event_runCheck()) {
+        return false;
+    }
+    return g_dComIfG_gameInfo.play.getEvent().checkHind(flag);
+}
```
lepelog, line 99: can they be moved to `d_com_inf_game.h`?

src/d/a/d_a_player.cpp
```diff
@@ -811,72 +666,74 @@ s16 daPy_py_c::checkNowWolfEyeUp() {
...
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-asm void daPy_py_c::forceRestartRoom(int param_0, u32 param_1, int param_2) {
-    nofralloc
-#include "asm/d/a/d_a_player/forceRestartRoom__9daPy_py_cFiUli.s"
+// move these inlines?
+inline BOOL i_dComIfGp_event_runCheck() {
+    return g_dComIfG_gameInfo.play.getEvent().runCheck();
+}
```
lepelog, line 672: If there isn't anything preventing that, I would move it to `d_com_inf_game.h`

src/d/meter/d_meter_map.cpp
TakaRikka: checkHind might be able to be moved, but some of the others have issues cause they have symbols in other files. specifically d_a_alink tends to un-inline a bunch of them, i assume cause its such a huge file, which causes re-def errors. so for now ive just been including those specific ones on a per-file basis

lepelog (approved):

PR #172, opened 2021-12-28
remove clang-format

lepelog:


notyourav (commented):
I get the removal of the files which enforce the requirement, but we should probably keep the .clang-format file so we can have at least some conventions to work off of.

TakaRikka:
i still feel like having it run on merges (with some tweaks) might be beneficial, and probably shouldnt break anything as long as handwritten asm is taken care of correctly. its not super important, but could still be nice for keeping some small things consistent

notyourav:
> i still feel like having it run on merges (with some tweaks) might be beneficial, and probably shouldnt break anything as long as handwritten asm is taken care of correctly. its not super important, but could still be nice for keeping some small things consistent

Yea we could have a blacklist.

lepelog:
I thought one of the reasons to remove clang format was because it makes the formatting overall worse, so it shouldn't be run at all.

We could also try to improve the formatting options: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

For example `AlignConsecutiveAssignments: true` would fix the enum formatting. It might be worth looking into cause in general I'd prefer to not have to worry about formatting myself.

Imo if we get rid of the clang-format-check, clang-format should not run at all, cause I can just run clang-format for those people that can't run clang-format.

TakaRikka:
yeah tweaking some of those options sounds like a good idea. and yeah i guess if someone is still gonna run clang on their end regardless, it doesnt need an action at all

notyourav:
> Imo if we get rid of the clang-format-check, clang-format should not run at all, cause I can just run clang-format for those people that can't run clang-format.

we should encourage code that follows the basic conventions so we dont just have a bunch of ugly pasted ghidra output. I think contributors should be able to use it at their discretion (avoiding problematic files?)

magcius:
> we should encourage code that follows the basic conventions

we should encourage any code. the standards are not particularly important, at least not yet.

TakaRikka (approved):

PR #176, opened 2022-01-12
J2DAnmLoader

lepelog:
All non matching because some dtors generate in the wrong place, the individual functions should match


include/JSystem/J2DGraph/J2DAnimation.h
```diff
@@ -10,197 +10,214 @@
...
 
+enum J2DAnmKind {
+    J2DAnmTransformKind = 0,
+    J2DAnmColorKind = 1,
+    J2DAnmTexPatternKind = 2,
+    J2DAnmTextureSRTKeyKind = 4,
+    J2DAnmTevRegKeyKind = 5,
+    J2DAnmVisibilityFullKind = 6,
+    J2DAnmVtxColorKind = 7
+};
```
TakaRikka, line 21: maybe formatting like `KIND_TRANSFORM` etc would be better? also all uppercase for enums is how i've been formatting them so far

TakaRikka (approved):

PR #177, opened 2022-01-16
Shiftability

jdflyer:
By changing (and adding) some tools we can generate a shiftable dol for the game.
extract_game_assets.py was changed to generate a file structure runnable by the dolphin emulator.
lcf.py was changed to generate symbols in the linker file that matches the shifted symbols.
package_game_assets.py was added to copy the assets over from game/ , copy rels over, and inject needed rels into a new RELS.arc

I also decompiled most of DynamicLink.cpp in the branch that I'm merging.
elf2dol was also updated to the latest version.


include/DynamicLink.h
```diff
@@ -2,5 +2,72 @@
...
+#include "m_Do/m_Do_dvd_thread.h"
+#include "JSystem/JKernel/JKRFileCache.h"
+
+struct DynamicModuleControlBase {
+    u16 mLinkCount; //0x0
+    u16 mDoLinkCount; //0x2
+    DynamicModuleControlBase* mPrev; //0x4
+    DynamicModuleControlBase* mNext; //0x8
+
+    /* 802621CC */ virtual ~DynamicModuleControlBase(); //0xC
```
TakaRikka, line 15: offsets should prob be formatted on the left side like `/* 0x0 */ u16 mLinkCount`
for the vtable, can just add a line under `mNext` like `/* 0xC */ // vtable`

include/DynamicLink.h
```diff
@@ -2,5 +2,72 @@
...
+    
+    OSModuleInfo* mModule; //0x10
+    void* mBss; //0x14
+    u32 unk_24; //0x18
+    const char* mName; //0x1c
+    u8 mResourceType; //0x20
+    u8 unk_33; //0x21
+    u16 mChecksum; //0x22
+    s32 mSize; //0x24
+    mDoDvdThd_callback_c* mAsyncLoadCallback; //0x28
```
TakaRikka, line 65: same comment about offset formatting

src/DynamicLink.cpp
```diff
@@ -360,210 +256,445 @@ asm void DynamicModuleControlBase::dump() {
...
-#pragma optimizewithasm off
-asm void ModuleUnresolved() {
-    nofralloc
-#include "asm/DynamicLink/ModuleUnresolved.s"
+extern "C" void ModuleUnresolved() {
+    OSReport_Error("\nError: リンクされていない関数が呼び出されました.\n");
+    OSReport_Error("Address:      Back Chain    LR Save\n");
+    u32 i = 0;
+    u32* stackPtr = (u32*)OSGetStackPointer();
+    while(stackPtr!=NULL&&(u32)stackPtr!=0xFFFFFFFF&&i++<0x10) {
```
TakaRikka, line 672: its probably better in general to add spaces between conditions and operators. kinda hard to read as a big block

TakaRikka (approved):

PR #183, opened 2022-03-20
Capitalization fixes & msl library folder names

Zen64:
Fixed capitalization of asm, include & libs files to better match frameworkF.map and to properly compile due to naming inconsistencies.

Merged MSL_C.PPCEABI.bare.H folder to msl_c folder in asm, include, and libs.


TakaRikka:
i think the msl_c in "msl_c/MSL_Common" should be capitalized like "MSL_C/MSL_Common"? other than that looks good

TakaRikka (approved):

PR #187, opened 2022-03-27
Add Progress.md

jdflyer:
This pull request adds Progress.md, a markdown file that has 3 tables to display the current progress of the project. I also updated the readme and contributing.md to clarify some things.


notyourav:
How does the progress file get updated?

jdflyer:
Every time tp.py is ran with the progress command it updates it.
(It should be ran with every pull request)

Pheenoh (approved):

PR #213, opened 2022-11-23
added VSCode devcontainer config files

kipcode66:
This adds configuration files for a VSCode devcontainer for people having trouble setting up the project and its dependencies (or people with Mx mac silicons).


RealMarioD (commented):
I don't think that's a useless command
When i was setting up the project for myself i had to give execute permissions to the executables or `./tp setup` would throw back a permission denied error

Pheenoh (approved):

PR #216, opened 2022-12-20
d_a_npc work + tp.py fixes

Pheenoh:
- Many matching and nonmatching in d_a_npc.cpp
- Fixed a bug that was causing some .inc files and .s files to be deleted erroneously when running the tp.py remove-unused-asm command


include/d/a/d_a_npc.h
```diff
@@ -111,6 +131,59 @@ struct daNpcT_motionAnmData_c {
...
+    /* 80145C74 */ int setPathInfo(u8, s8, u8);
+    /* 80145D2C */ int setNextPathInfo(s8, u8);
+    /* 80145DA0 */ void reverse();
+    /* 80145DD0 */ int setNextIdx(int);
+    /* 80145E38 */ int getDstPos(cXyz, cXyz*, int);
+    /* 80145FB4 */ int getDstPosH(cXyz, cXyz*, int, int);
+    /* 80146188 */ int chkPassed1(cXyz, int);
+    /* 801464D8 */ int chkPassed2(cXyz, cXyz*, int, int);
+
+    inline cXyz* getPntPos(u16 i_idx) {
```
TakaRikka, line 167: just a couple notes, but its not anything super important. inline keyword isn't needed if you're defining it in the class afaik

include/d/d_path.h
```diff
@@ -14,6 +15,9 @@ struct dPath {
     /* 0x8 */ dStage_dPnt_c* m_points;
 };
 
+// should be move to dlib TU per chn debug
```
TakaRikka, line 18: if this is a weak inline it prob doesn't actually belong to the d_lib TU, it just gets emitted there by the compiler. this can be seen in retail too with stuff like daPy_py_c inlines being in other unrelated TUs

src/d/a/d_a_npc.cpp
```diff
@@ -913,142 +768,365 @@ s32 daNpcT_MotionSeqMngr_c::checkEndSequence() {
...
-#pragma optimizewithasm off
-asm void daNpcT_Path_c::setNextPathInfo(s8 param_0, u8 param_1) {
-    nofralloc
-#include "asm/d/a/d_a_npc/setNextPathInfo__13daNpcT_Path_cFScUc.s"
+int daNpcT_Path_c::setNextPathInfo(s8 param_0, u8 i_direction) {
+    mIdx = 0;
+    mDirection = i_direction;
+    u16 next_id = mpRoomPath->m_nextID;
+
+    if (next_id != 65535) {
```
TakaRikka, line 901: this comparison probably looks better as 0xFFFF rather than 65535

src/d/a/d_a_npc.cpp
```diff
@@ -1969,16 +2121,73 @@ asm void daNpcT_c::ctrlMotion() {
...
+                mSubMtd = (profile_method_class*)-1;
+            } else if (actor_command == 6) {
+                if (talk_partner->mSubMtd != mSubMtd) {
+                    *tmp0 = dComIfGp_getMesgFaceAnimeAttrInfo();
+                    *tmp1 = dComIfGp_getMesgAnimeAttrInfo();
+                    mSubMtd = talk_partner->mSubMtd;
+                }
+
+                if (talk_partner_conv->isMouthCheck()) {
+                    field_0xd98 &= 0xffffbfff;
```
TakaRikka, line 2155: these &= 0xFFFFBFFF and similar patterns are usually something like &= ~0x4000

src/d/a/d_a_npc.cpp
```diff
@@ -3351,24 +3835,15 @@ bool daBaseNpc_c::dbgDraw() {
...
  */
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-asm daBaseNpc_moveBgActor_c::daBaseNpc_moveBgActor_c() {
-    nofralloc
-#include "asm/d/a/d_a_npc/__ct__23daBaseNpc_moveBgActor_cFv.s"
+// this probably has a more correct solution
+daBaseNpc_moveBgActor_c::daBaseNpc_moveBgActor_c() {
+    setVtable(&__vt__23daBaseNpc_moveBgActor_c);
```
TakaRikka, line 3840: vtable setting is compiler generated after you setup the virtual functions in the class def

TakaRikka (approved):

PR #223, opened 2022-12-27
d_ky_thunder first pass

Pheenoh:


src/d/d_ky_thunder.cpp
```diff
@@ -243,9 +214,26 @@ static asm void dThunder_Delete(dThunder_c* param_0) {
 #include "asm/d/d_ky_thunder/dThunder_Delete__FP10dThunder_c.s"
 }
 #pragma pop
+#endif
 
 /* 801AE3FC-801AE458 1A8D3C 005C+00 1/0 0/0 0/0 .text            dThunder_Create__FP12kankyo_class
  */
+#ifdef NONMATCHING
+// regalloc
+static int dThunder_Create(kankyo_class* param_0) {
```
TakaRikka, line 223: just noting that the main method functions of actors should probably be returing cPhs enums i think, but i haven't been very consistent about it either. its not a big deal rn, but eventually we should switch all actor method returns to cPhs enum

TakaRikka (approved):

src/d/d_ky_thunder.cpp
Pheenoh: will merge for now but let's rework these at some point

PR #224, opened 2022-12-27
f_op_actor matches

Pheenoh:


src/f_op/f_op_actor.cpp
```diff
@@ -371,36 +360,40 @@ static asm int fopAc_Create(void* param_0) {
...
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-asm void fopEn_enemy_c::initBallModel() {
-    nofralloc
-#include "asm/f_op/f_op_actor/initBallModel__13fopEn_enemy_cFv.s"
-}
-#pragma pop
+bool fopEn_enemy_c::initBallModel() {
+    void* objRes = dComIfG_getObjectRes("Alink",((daAlink_c*)this)->getBallModelIdx());
```
TakaRikka, line 369: these `((daAlink_c*)this)` accesses seem fake, these `getBall` functions might be static instead? so accessed like `daAlink_c::getBallModelIdx()` and such

TakaRikka (approved):

PR #230, opened 2022-12-31
Script To Package built game into ISO

jdflyer:
This PR includes a python script that takes the build/dolzel2/game/ folder and creates an ISO image out of it. While the built ISO works on hardware, it isn't identical to the source ISO. The filesystem layout is identical other than the size of rels being different due to a different compression algorithm from the source rels. Also, I'm not sure about how the files should be laid out on the disk; it seems almost random in the source ISO. The script currently just places files on the disc how they are laid out in the filesystem (with 0x100 padding between them).


TakaRikka:
guessing there's no idea about the original compression algorithm? it's not a major deal as long as its functionally the same, but would be nice to get them fully matching one day

TakaRikka (approved):

barzamin (commented):
can you blacken or otherwise format the `packageISO` script? it's a little bit gungy at the moment

barzamin (approved):

PR #239, opened 2023-01-03
Add frank.py

jdflyer:
This adds frank.py (from the pikmin 2 decomp) which fixes epilogue issues in the 1.2.5 compiler that the os libs use.


TakaRikka:
is this using the 1.2.5e compiler or just base 1.2.5?

jdflyer:
It's just frank.py on object files compiled with base 1.2.5

TakaRikka:
ah okay, wasn't sure if it was required or not. seems fine then

jdflyer:
Let me see if I can get rid of any more diffs by using 1.2.5e

jdflyer:
It seems like 1.2.5e needs some setup to compile files so I guess we can just leave it on base 1.2.5 for now

TakaRikka (approved):

PR #241, opened 2023-01-05
Add header dependencies to Makefiles

Jcw87:
For smarter rebuilds.

This needs testing in more environments, like Linux and WSL


jdflyer:
I can confirm this works on linux. However, I have some concerns about build time. A Full build on my system on the current repository takes 6 seconds less than this build. Additionally, make on this PR takes about twice the time to process files before execution than the current repository (this time a makes a huge difference when using diff.py). Is there any way to make including dependency files optional?

Jcw87:
I've added a `DISABLE_DEPS` makefile variable that will ignore all extra processing related to the header dependencies if it is set to anything. This includes the compiler generating the `.d` file, the post-processing step to fix the path separators, and the include statement that uses the `.d` files. `diff.py` still needs to be changed to use it.

jdflyer:
Looks good! If you add DISABLE_DEPS=1 to the makeflags in diff_settings.py it's good with me!

jdflyer (approved):

Jcw87:
I got so focused on handling the 800 or so makefiles for the libraries and rels that I forgot about main.dol. Should probably give that a quick review before merging

PR #247, opened 2023-01-19
Add 1.2.5e to several makefiles and matching dolphin lib stuff

hatal175:
- Adding 1.2.5e requirement to tp.py and readme.
- Use on dolphin libs ar,ai and card. Still more to go.
- Most matches are from metroid prime deomp + minor changes needed.
- Minor improvement/fix to diff.py.
- Required fake inline versions of some functions (see discord). Could probably be fixed with compiler flags. I'll try looking when I have time but this pull request is still valid in my opinion.


hatal175:
I'll fix this tomorrow, not sure what it wants.

Pheenoh (approved):

PR #251, opened 2023-01-22
rename mwcceppc_patched.exe to mwcceppc_modded.exe for better windows compatibility

Jcw87:
Windows treats any executable with `patch` in the name as an installer, and will try to request admin permissions through UAC.
This fixes permission denied errors in mingw.


hatal175:
Missed a spot in the github check

TakaRikka (approved):

PR #257, opened 2023-01-27
fake the -MP compiler option

Jcw87:
There is an issue where the build will get blocked if any include gets deleted (like headers or asm). This is normally resolved with the `-MP` flag, which instructs the compiler to generate phony make targets for each include. However, the metrowerks compiler doesn't support this option, so I made `transform-deps.py` do it instead.

### DIFF
--- a/tools/transform-dep.py
+++ b/tools/transform-dep.py
@@ -3,56 +3,66 @@
 import argparse
 import os
 from platform import uname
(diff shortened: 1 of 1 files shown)


barzamin (approved):
this looks good to me on cursory inspection; assuming build passes and you've tested, lgtm.

PR #263, opened 2023-01-30
Setup fixes

Jcw87:
This PR addresses the following issues:

requirements.txt had 2 entries which were standard python libraries and could not be installed by pip
`make tools` failed until the user created the build directory manually
WSL users wouldn't be able to run the compilers without using `chmod +x` first
`./tp setup` did not copy LMGR326B.dll for 1.2.5e

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,7 +152,7 @@ clean_rels:
 	rm -f -d -r $(BUILD_DIR)/rel
 	rm -f $(BUILD_PATH)/*.rel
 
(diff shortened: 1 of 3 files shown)


jdflyer:
We should probably add make tools into the setup script to remove a step from setup

Jcw87:
I'm actually in the middle of doing that right now.

Jcw87:
I had to delay-load a couple of imports to make this work, or the script would just exit immediately.

jdflyer (approved):

PR #266, opened 2023-02-01
Dvdlow and others

hatal175:
- dvd low work
- Fixed some time related nonmatching (problematic macro)
- misc nonmatching fixes
- diff now prints stderr when it fails mid work so you don't have to exit and fix


TakaRikka:
are you sure about these changes to dKyw_get_AllWind_vecpow? the original map file from in the game says the is `dKyw_get_AllWind_vecpow__FP4cXyz`

TakaRikka:
hm maybe should get some other opinions on it, not sure the best way to handle that. not sure if the maps could've somehow been generated with errors or something

Julgodis:
I think symbol map is correct and that the `r3` argument is generated by the compiler because `dKyw_get_AllWind_vecpow__FP4cXyz` returns a `cXyz` by value. Every location that is calling `dKyw_get_AllWind_vecpow__FP4cXyz` will pass in a pointer to a stack allocated `cXyz` object that the function will fill out.

hatal175:
Let me try doing that.

jdflyer (approved):

src/d/d_insect.cpp
```diff
@@ -15,38 +15,21 @@
...
-extern "C" void newData__13dDlst_peekZ_cFssPUl();
 extern "C" void Insect_Release__9dInsect_cFv();
-extern "C" void isCatchNotGiveInsect__14dMenu_Insect_cFUc();
-extern "C" u8 saveBitLabels__16dSv_event_flag_c[1644 + 4 /* padding */];
 
 //
 // Declarations:
 //
 
+inline int dComIfGs_isEventBit(u16 event) {
```
TakaRikka, line 28: is this inline needed here or does the `i_` version in the d_com_inf_game.h header work?

hatal175:
Tried it but unfortunately it's needed. Didvit after I saw  it in many
places.

PR #270, opened 2023-02-05
d_a_crod almost fully OK

TakaRikka:
only posMove left with just 1 small issue


include/d/a/d_a_player.h
```diff
@@ -708,6 +711,9 @@ class daPy_py_c : public fopAc_ac_c {
         }
     }
 
+    s16 getBodyAngleX() const { return mLookAngleY; }
+    s16 getBodyAngleY() const { return field_0x59e; }
```
hatal175, line 715: Rename field_0x59e?

rel/d/a/d_a_crod/d_a_crod.cpp
```diff
@@ -1,324 +1,182 @@
...
+
+    if (!fopAcM_entrySolidHeap(this, daCrod_createHeap, 0x10C0)) {
+        return cPhs_ERROR_e;
+    }
+
+    mCcStts.Init(10, 0, this);
+    mAtCps.Set(l_atCpsSrc);
+    mAtCps.SetStts(&mCcStts);
+
+    field_0x524 = mpBallModel;
```
hatal175, line 101: Rename field_0x524?

src/d/a/d_a_alink.cpp
```diff
@@ -18733,9 +18733,14 @@ extern "C" asm void __dt__9cCcD_SttsFv() {
...
 
 /* 801410EC-8014111C 13BA2C 0030+00 1/1 0/0 0/0 .text            __ct__16daPy_actorKeep_cFv */
-daPy_actorKeep_c::daPy_actorKeep_c() {
-    clearData();
+#pragma push
+#pragma optimization_level 0
+#pragma optimizewithasm off
+extern "C" asm void __ct__16daPy_actorKeep_cFv() {
+    nofralloc
+#include "asm/d/a/d_a_alink/__ct__16daPy_actorKeep_cFv.s"
```
hatal175, line 18741: Why extract to an assembly file? If for a good reason, maybe leave the c code with a NONMATCHING ifdef?

TakaRikka: this is an inlined constructor, so this is supposed to be a compiler generated symbol once the main daAlink_c constructor is matching. i still need the inlined constructor to match other classes that use daPy_actorKeep_c in the meantime tho

hatal175 (approved):

jdflyer (approved):

PR #273, opened 2023-02-07
most of d_a_tag_spinner decompiled

Trueffeloot:
Last function is **daTagSpinner_c::execute**. Feel free to finish it up!


include/rel/d/a/tag/d_a_tag_spinner/d_a_tag_spinner.h
```diff
@@ -1,6 +1,31 @@
 #ifndef D_A_TAG_SPINNER_H
 #define D_A_TAG_SPINNER_H
 
+#include "d/com/d_com_inf_game.h"
+#include "dol2asm.h"
```
TakaRikka, line 5: if `dol2asm.h` isnt required for building this rel anymore, i'd remove the include here. the `dolphin/types.h` include is also unnecessary here since it's included in a bunch of other sub headers here

include/rel/d/a/tag/d_a_tag_spinner/d_a_tag_spinner.h
```diff
@@ -1,6 +1,31 @@
...
+    /* 80D618C4 */ void setBaseMtx();
+    /* 80D6191C */ int Create();
+    /* 80D6196C */ int CreateHeap();
+    /* 80D61974 */ int create();
+    /* 80D61A00 */ int execute();
+    /* 80D61AB4 */ int _delete();
+
+    /* 0x568 */ f32 field_0x568;
+    /* 0x56c */ f32 field_0x56c;
+    inline f32 GetR() { return field_0x568; }
```
TakaRikka, line 22: based on the inline name, i would call `field_0x568` something like mRadius. the inline keyword is also unnecessary if you're defining the function inside the class. (also just personal preference but i prefer having the inlines above the class members)

rel/d/a/tag/d_a_tag_spinner/d_a_tag_spinner.cpp
```diff
@@ -63,177 +34,110 @@ extern "C" void fopAcIt_Judge__FPFPvPv_PvPv();
...
+int daTagSpinner_c::create() {
+    int iVar1;
+    int iVar2;
+    if (!fopAcM_CheckCondition(this, 8)) {
+        new (this) daTagSpinner_c();
+        fopAcM_OnCondition(this, 8);
+    }
+
+    if (!fopAcM_entrySolidHeap(this, CheckCreateHeap, 0)) {
+        iVar2 = 5;
```
TakaRikka, line 99: can this section be rewritten with returns instead of temp variables?
something like
```c++
if (!fopAcM_entrySolidHeap(this, CheckCreateHeap, 0)) {
    return 5;
}

if (!Create()) {
    return 5;
}

return 4;
```

Trueffeloot: Doesn't match unfortunately. I think I have to stay with these temp variables

include/rel/d/a/tag/d_a_tag_spinner/d_a_tag_spinner.h
Trueffeloot: Did you mean with "dolphin/types.h" "d/com/d_com_inf_game.h"? If that's the case, I removed it. "dol2asm.h" is still required for the last function "daTagSpinner_c::execute"

hatal175:
Why does it say there are +5000 files changed? I can't even see the commit because github gives up.

TakaRikka:
looks like almost every file in the project was touched though without updating any of the contents. i'd probably revert the last commit and try to reset to a stable point before those file changes

Trueffeloot:
Yeah I'm sorry don't know what VSCode did there

Pheenoh:
Yeah I also can't view it, @Trueffeloot you'll have to fix whatever happened here before we can merge

Pheenoh:
closed per request

PR #288, opened 2023-02-16
d_a_alink_swim work / fix JASGlobalInstance

TakaRikka:
most of d_a_alink_swim done. JASGlobalInstance sInstance data set up by using relative offsets to the data. it's a tradeoff, but the benefits outweigh the negatives imo. some Z2 cleanup in the process


Pheenoh (approved):

Pheenoh:
This should be fine for now, but let's keep track somewhere of the exceptions we're making here. We may want to implement a preprocessing script in the future to auto determine these offsets so no shiftability issue pops up

TakaRikka:
yeah it should only cause problems if you insert new data between the sInstance data and the __OSReport_disable symbol, so it's a pretty small issue thats unlikely to be run into, but needs to be noted

PR #292, opened 2023-02-17
d_a_tag_shop_camera OK, d_a_tag_watchge OK, JUTFader OK, JUtility work from Pikmin 2

Pheenoh:


libs/JSystem/JUtility/JUTResFont.cpp
```diff
@@ -309,6 +324,89 @@ SECTION_SDATA2 static f64 lit_651 = 4503599627370496.0 /* cast u32 to float */;
...
+	u16 sumHeight              = t_height + used_glyphs->cellHeight;
+	s32 width_ratio            = shift_width / tex_width;
+	s32 height_ratio           = shift_height / tex_height;
+
+	GXSetVtxAttrFmt(GX_VTXFMT0, GX_VA_POS, GX_POS_XYZ, GX_F32, 0);
+	GXBegin(GX_QUADS, GX_VTXFMT0, 4);
+
+	f32 zero_f = 0.0f;
+
+	HW_REG(0xCC008000, f32) = width_val;
```
TakaRikka, line 377: the HW_REG macro in general should be removed because the debug rom shows the inlines they had for writing to the GXFIFO, such as GXPosition3f32 and GXTexCoord2u16

libs/JSystem/JUtility/JUTResFont.cpp
```diff
@@ -396,217 +487,81 @@ asm u16 JUTResFont::getCellHeight() const {
...
-    0x62,
-    0x81,
-    0x70,
-    0x81,
-    0x60,
-    /* padding */
-    0x00,
-    0x00,
+static const u16 halftofull[95] = {
+		0x8140, 0x8149, 0x8168, 0x8194, 0x8190, 0x8193, 0x8195, 0x8166, 0x8169, 0x816A, 0x8196, 0x817B, 0x8143, 0x817C, 0x8144, 0x815E,
```
TakaRikka, line 502: running clang might be good to clean up formatting a bit

libs/JSystem/JUtility/JUTResFont.cpp
Pheenoh: ah good point, i just copy pasta'd this from pikmin. will update

Pheenoh: GXTexCoord2u16 inline is defined but commented out, it still has issues that will need to be fixed at some point. It caused relocation issues due to it being referenced in d_a_movie_player rel. Didn't spend much time on it but need to figure out why that's happening at some point

libs/JSystem/JUtility/JUTResFont.cpp
Pheenoh: clang format didn't do anything here so i did some manual shifting around to make it look a bit cleaner

libs/JSystem/JUtility/JUTResFont.cpp
TakaRikka: im guessing its probably a weak function like other inline issues we've had, so might need to do the `i_` thing for now until we're able to properly fix it. but not a super big issue atm

TakaRikka (approved):

PR #302, opened 2023-03-01
cleanup JSystem files / d_a_alink_hang first pass

TakaRikka:
25% dol code


libs/JSystem/JKernel/JKRExpHeap.cpp
```diff
@@ -1067,6 +989,7 @@ asm bool JKRExpHeap::dump() {
 #endif
 
 /* 802D05CC-802D0810 2CAF0C 0244+00 1/0 0/0 0/0 .text            dump_sort__10JKRExpHeapFv */
+// using r0 instead of r4
```
jdflyer, line 992: You could try putting whatever is in r0 as the second parameter for the print function

hatal175 (approved):

PR #303, opened 2023-03-01
d_a_tag_rmbit_sw: OK

Trueffeloot:
With that, I added the definition for dComIfGs_offVisitedRoom in d_com_inf_game.h


Pheenoh (approved):
lgtm

Trueffeloot:
I could give the fields better names tho. Do you know what I could call them? They are used in the init function and are assigned to mScale multiplied with some numbers

PR #304, opened 2023-03-03
Some work on d_menu_fishing

Trueffeloot:
First time working on something different than the tag files. Hopefully this is good to go


include/d/menu/d_menu_fishing.h
```diff
@@ -5,9 +5,12 @@
 #include "d/d_lib.h"
 #include "d/d_select_cursor.h"
 #include "dolphin/types.h"
+#include "m_Do/m_Do_graphic.h"
+#include "m_Do/m_Do_dvd_thread.h"
+#include "m_Do/m_Do_controller_pad.h"
+#include "d/msg/d_msg_string.h"
```
TakaRikka, line 11: if these are only needed for the cpp file and not for anything in the header, it's better to only include them in the cpp file. including them in the header unnecessarily creates unnecessary dependencies if any other file includes it

src/d/menu/d_menu_fishing.cpp
```diff
@@ -178,17 +166,17 @@ asm dMenu_Fishing_c::~dMenu_Fishing_c() {
...
 /* 801C504C-801C50B4 1BF98C 0068+00 1/1 0/0 0/0 .text            _create__15dMenu_Fishing_cFv */
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-asm void dMenu_Fishing_c::_create() {
-    nofralloc
-#include "asm/d/menu/d_menu_fishing/_create__15dMenu_Fishing_cFv.s"
+void dMenu_Fishing_c::_create() {
+    dMsgString_c* dMsgString_c_ptr = new dMsgString_c;
+    mpString = dMsgString_c_ptr;
```
TakaRikka, line 174: this can be optimized to just `mpString = new dMsgString_c();`

src/d/menu/d_menu_fishing.cpp
```diff
@@ -216,6 +204,18 @@ SECTION_SDATA2 static f32 lit_3905 = 608.0f;
 SECTION_SDATA2 static f32 lit_3906 = 448.0f;
 
 /* 801C514C-801C5204 1BFA8C 00B8+00 1/1 1/1 0/0 .text            _draw__15dMenu_Fishing_cFv */
+#ifdef NONMATCHING
+// matches without literals
+void dMenu_Fishing_c::_draw() {
+    if (field_0x8) {
+        J2DGrafContext* pJVar1 = g_dComIfG_gameInfo.play.mCurrentGrafPort;
+        mpBlackTex->setAlpha(0xff);
+        mpBlackTex->draw(0,0,0x4083000000000000,0x407c000000000000,0,0,0);
```
TakaRikka, line 213: i assume those long hex strings are supposed to be float literals, which should replace those

src/d/menu/d_menu_fishing.cpp
```diff
@@ -216,6 +204,18 @@ SECTION_SDATA2 static f32 lit_3905 = 608.0f;
 SECTION_SDATA2 static f32 lit_3906 = 448.0f;
 
 /* 801C514C-801C5204 1BFA8C 00B8+00 1/1 1/1 0/0 .text            _draw__15dMenu_Fishing_cFv */
+#ifdef NONMATCHING
+// matches without literals
+void dMenu_Fishing_c::_draw() {
+    if (field_0x8) {
+        J2DGrafContext* pJVar1 = g_dComIfG_gameInfo.play.mCurrentGrafPort;
```
TakaRikka, line 211: the `dComIfGp_getCurrentGrafPort()` inline should be used instead of directly accessing it if possible

src/d/menu/d_menu_fishing.cpp
```diff
@@ -524,14 +516,19 @@ asm void dMenu_Fishing_c::setBButtonString(u16 param_0) {
 #pragma pop
 
 /* 801C6018-801C605C 1C0958 0044+00 1/1 0/0 0/0 .text            getFigure__15dMenu_Fishing_cFi */
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-asm void dMenu_Fishing_c::getFigure(int param_0) {
-    nofralloc
-#include "asm/d/menu/d_menu_fishing/getFigure__15dMenu_Fishing_cFi.s"
+int dMenu_Fishing_c::getFigure(int i_this) {
```
TakaRikka, line 519: i don't this `i_this` makes sense for a parameter name here, as that name is used for C-style object passing. not exactly sure what the param is atm though, so probably just leave it as `param_0` for now

src/d/menu/d_menu_fishing.cpp
```diff
@@ -161,6 +135,19 @@ SECTION_DATA extern void* __vt__15dMenu_Fishing_c[4 + 3 /* padding */] = {
 
 /* 801C4D54-801C4D98 1BF694 0044+00 0/0 2/2 0/0 .text
  * __ct__15dMenu_Fishing_cFP10JKRExpHeapP9STControlP10CSTControl */
+#ifdef NONMATCHING
+dMenu_Fishing_c::dMenu_Fishing_c(JKRExpHeap* heap, STControl* stControl, CSTControl* cstControl) {
```
jdflyer, line 139: Check debug on the constructor because I remember there's some inheritance going on when I last looked at it

src/d/menu/d_menu_fishing.cpp
Trueffeloot: hmm I can't match it like this. There is a  **lis     r3,g_dComIfG_gameInfo@ha** instruction missing at the end and later on, where the pointer gets stored as a parameter,  I'm getting a nonmatching saying that this **lis     r3,g_dComIfG_gameInfo@ha**  instruction is wrong and that I have to instead move a register.
So I assume this mr is the pointer being stored as a parameter in mpScreen and mpIconScreen

Trueffeloot: Nvm I misunderstood your comment, I had to assign this inline directly to the pointer variable! Sorry! :D

src/d/menu/d_menu_fishing.cpp
Trueffeloot: I'm not knowledged enough for that rn. The constructors of a class can also inherit? The class itself is already inheriting from dDlst_base_c. Idk what to do with that inheritance in the constructor

PR #306, opened 2023-03-07
Misc Changes

Pheenoh:
- d_a_tag_mstop work
- d_event_debug flag table added. dumped and translated from the debug rom
- setup enums for saveBitLabels and tempBitLabels from event flag table
- update some enum refs to saveBitLabels (there are still many left)
- cleanup headers/comments on f_pc TUs
- closer match on remaining function in d_save
- add back in function comments to d_save
- setup a few scene_process_profile_defintions
- cleanup and add Lunarsoap's d_msg matches


TakaRikka:
overall looks fine, but just wondering whether `d_event_debug.inc` should just be documentation in `docs/` or if we want to start implementing debug rom stuff here so that a GC debug build can eventually be built

Pheenoh:
I just went ahead and setup the tables in d_event with a debug flag check (similar to what was done in a handful of JSystem files)

TakaRikka (approved):

TakaRikka:
yeah we should come back to it once we decide how to handle multi version

PR #309, opened 2023-03-18
d_file_select, JASResArcLoader, d_file_select, d_a_obj_stone, d_a_obj_stone_mark

Pheenoh:


src/d/file/d_file_select.cpp
```diff
@@ -2734,8 +2289,33 @@ asm void dFile_select_c::menuSelectCansel() {
 #include "asm/d/file/d_file_select/menuSelectCansel__14dFile_select_cFv.s"
 }
 #pragma pop
+#endif
 
 /* 80186B48-80186C84 181488 013C+00 8/8 0/0 0/0 .text menuMoveAnmInitSet__14dFile_select_cFii */
+#ifdef NONMATCHING
+// getPanePtr() calling function past end of vtable?
```
TakaRikka, line 2296: getPanePtr seems to return a pointer to the base J2DPane of a pane object so that it can be used for any pane object type afaict, so it just gets casted into whatever it's supposed to be, like J2DTextBox, or J2DPicture, etc

TakaRikka (approved):

src/d/file/d_file_select.cpp
Pheenoh: Good info. Will play around with other J2D* classes

PR #314, opened 2023-04-05
d_a_tag_Lantern OK

Trueffeloot:


rel/d/a/tag/d_a_tag_lantern/d_a_tag_lantern.cpp
```diff
@@ -1,298 +1,145 @@
...
 /* 8048EDDC-8048EF6C 00021C 0190+00 2/2 0/0 0/0 .text            Execute__15daTag_Lantern_cFv */
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-asm void daTag_Lantern_c::Execute() {
-    nofralloc
-#include "asm/rel/d/a/tag/d_a_tag_lantern/d_a_tag_lantern/Execute__15daTag_Lantern_cFv.s"
+int daTag_Lantern_c::Execute() {
+    s32 sVar;
+    dComIfG_play_c& play = g_dComIfG_gameInfo.getPlay();
```
TakaRikka, line 30: is this temp necessary? debug makes it seem like this should just be `dComIfGp_event_reset` on line 37

TakaRikka (commented):
just 1 minor thing, otherwise lgtm

rel/d/a/tag/d_a_tag_lantern/d_a_tag_lantern.cpp
Trueffeloot: it is. I used to use that inline while working on that function but it didn't match (inline generated wrong asm instructions). For some reason, this temp is currently necessary

TakaRikka (approved):

PR #315, opened 2023-04-07
d_a_obj_cowdoor OK

Trueffeloot:
Also had to define a function in the d_bg_s header for d_a_obj_cowdoor and I added the descriptions in the actor_process_profile_definition of d_a_obj_bombf because I forgot it last time


rel/d/a/obj/d_a_obj_tmoon/d_a_obj_tmoon.cpp
```diff
@@ -1,273 +1,140 @@
...
-/* ############################################################################################## */
-/* 80D12F98-80D12F98 000000 0000+00 0/0 0/0 0/0 .rodata          @stringBase0 */
-#pragma push
-#pragma force_active on
-SECTION_DEAD static char const* const stringBase_80D12F98 = "A_TMoon";
-#pragma pop
 
 /* 80D12FA0-80D12FA4 -00001 0004+00 3/3 0/0 0/0 .data            l_arcName */
-SECTION_DATA static void* l_arcName = (void*)&d_a_obj_tmoon__stringBase0;
+static const char* l_arcName[] = {"A_TMoon"};
```
TakaRikka, line 36: this applies to all instances of this throughout the PR, but if it's a `const char*` to just 1 string, you don't need to use `[]` since its not an array. you should be able to define it like `static const char* l_arcName = "A_TMoon";`.

this should make it so that when you use it in a function, you can just use it like `dComIfG_getObjectRes(l_arcName, 3)` instead of `l_arcName[0]`

magcius: Those two are different and will have different asm -- the array is needed to match the extra dereference.

Trueffeloot: It actually matches as just a static const char* in both files as Taka meant. So there should be no problem

hatal175 (approved):

PR #317, opened 2023-04-19
add MSL_C/stdio.h

Jcw87:


TakaRikka:
should we have MSL_C headers in a way that that we can include them like `#include <stdio.h>` ? same with any other library header so that it's closer to how it would be in a normal C/C++ project

Jcw87:
I'm fine with the way it is now. My main goal with this PR was to replace `&__files.stdout` in the game code with `stdout` for better compatibility with other compilers.

TakaRikka (approved):

PR #322, opened 2023-04-23
d_a_obj_zcloth WIP

Trueffeloot:
All functions (except the compiler-generated dtors which get deleted) match. The assembly is really weird when diffing the create function tho. I didn't know what to do so I just ifdeffed it out for now.
Also matched two functions in d_menu_fishing.cpp (which I just copied from d_menu_calibration)


TakaRikka:
pull from master to fix the progress conflict and then it should be good to go

Trueffeloot:
progress now updated. pr should be good to go now

hatal175 (approved):

PR #328, opened 2023-05-06
Do some work on f_op_actor_mng

hatal175:
- Also made decompctx work better with some of the characters


include/rel/d/a/tag/d_a_tag_stream/d_a_tag_stream.h
```diff
@@ -11,17 +11,38 @@ class daTagStream_c : public fopAc_ac_c {
         return mPriority;
     }
 
+    inline static daTagStream_c* getTop() {
```
TakaRikka, line 14: i dont think the inline keyword is needed if your defining the function in the class

TakaRikka (commented):
just some small things, otherwise looks good

TakaRikka (approved):

PR #329, opened 2023-05-09
J3DJoint OK

hatal175:
Curious copy in J3DJoint::J3DJoint. If anyone ran into something similar, would be nice to know.


libs/JSystem/J3DGraphAnimator/J3DJoint.cpp
```diff
@@ -235,46 +200,105 @@ J3DJoint::J3DJoint() {
     mJntNo = 0;
     mKind = 1;
     mScaleCompensate = false;
-    mTransformInfo = j3dDefaultTransformInfo;
+    *(copyhelper*)&mTransformInfo = *(copyhelper*)&j3dDefaultTransformInfo;
```
TakaRikka, line 203: a comment about this being a fake match would probably be good just incase we can clean it up later on, since i doubt something like this was actually done

TakaRikka (approved):

PR #334, opened 2023-05-11
Remove #includes from headers

Jcw87:
This PR aims to reduce the number of files that need to be recompiled when you edit a header. This is done by removing #includes from headers than can simply be replaced with a few forward declarations.

If you think I went too far, I can dial it back if need be, or split this into smaller PRs


TakaRikka (commented):
i'd like another opinion on this, but as far i can tell this looks good to me. future work should follow the same style of forward declaring where possible

hatal175 (approved):
Looks good to me

Jcw87:
Updated to account for the other PR that was recently merged

PR #339, opened 2023-05-27
setFadeColor fix, some d_msg_flow functions

hatal175:


rel/d/a/d_a_title/d_a_title_nonmatching.cpp
```diff
@@ -468,25 +475,13 @@ void daTitle_c::nextScene_init() {
 }
 
 /* 80D67550-80D675EC 000B30 009C+00 1/0 0/0 0/0 .text            nextScene_proc__9daTitle_cFv */
-// setFadeColor store order issue
-#ifdef NONMATCHING
 void daTitle_c::nextScene_proc() {
     if (!fopOvlpM_IsPeek() && !mDoRst::isReset()) {
         scene_class* playScene = fopScnM_SearchByID(dStage_roomControl_c::getProcID());
         fopScnM_ChangeReq(playScene, 13, 0, 5);
         mDoGph_gInf_c::setFadeColor(*(JUtility::TColor*)&g_blackColor);
```
TakaRikka, line 482: did the setFadeColor fix allow for removing these type of ugly casts? its fine if not, but eventually a fix for it should be figured out

TakaRikka (approved):

rel/d/a/d_a_title/d_a_title_nonmatching.cpp
hatal175: Seems like it needs the TColor for the copy ctor. Maybe if you can find a way to define g_blackColor as TColor and still have it on the same data section?

PR #350, opened 2023-06-19
TP Progress step for displaying

Trueffeloot:
thanks to pheenoh for creating the github actions!


Pheenoh (approved):

hatal175:
Should you delete the progress file and add a widget to the readme or is this for another PR?

Trueffeloot:
@hatal175 didn't think about that rn. For now I'd prefer still keeping the progress.md file because you can see all the single rels too there. And about a json-shield widget, we might add that later

PR #356, opened 2023-06-28
Work on d_select_cursor and d_com_static

hatal175:
Also some files deleted due to #355


rel/d/a/d_a_talk/d_a_talk.cpp
```diff
@@ -79,10 +44,9 @@ int daTalk_c::create() {
 }
 
 /* 80D664AC-80D6665C 0001AC 01B0+00 1/1 0/0 0/0 .text            execute__8daTalk_cFv */
-// g_dComIfG_gameInfo.play.mEvent being reloaded when it shouldnt be
-#ifdef NONMATCHING
 int daTalk_c::execute() {
-    if (i_dComIfGp_event_runCheck()) {
+    dComIfG_inf_c& dcomif = g_dComIfG_gameInfo;
```
TakaRikka, line 48: feel like this is prob a fake match but o well, figuring out these loads is annoying

TakaRikka (approved):

PR #359, opened 2023-07-03
d_cc_d_nonmatching

Jcw87:
Changes `d_cc_d_nonmatching.cpp` to use inlines from debug rom. Also adds a match for `GObjInf::ChkAtNoGuard`


include/SSystem/SComponent/c_cc_d.h
```diff
@@ -152,13 +163,15 @@ class cCcD_CpsAttr : public cCcD_ShapeAttr, public cM3dGCps {
 
 STATIC_ASSERT(0x40 == sizeof(cCcD_CpsAttr));
 
-class cCcD_SrcSphAttr : public cM3dGSphS {};
+struct cCcD_SrcSphAttr {
+    cM3dGSphS mSph;
+};
```
TakaRikka, line 168: do these not work inheritance?

Jcw87: I couldn't get the data initializers to work. If it doesn't use inheritance, adding more curly braces makes it work. If you can think of a way to make it work, I'll change it.

hatal175 (approved):

PR #372, opened 2023-07-16
Use compiler 1.2.5n instead of frank

hatal175:


hatal175:
@Pheenoh Can you update your image with the latest compiler dir?

Pheenoh:
> @Pheenoh Can you update your image with the latest compiler dir?

Done. Although there appears to be another issue. But the latest compilers should be on there now -- I updated the container to use the ones from the discord link you updated.

hatal175:
This is very peculiar. I can't reproduce this in my repo. It's not even in a file in the dolphin lib.
Can someone do me a favor and clone this branch?

jdflyer:
Does the new compiler directory have the patched version of the 2.7 compiler?

hatal175:
Good point. The setup script does that.

Pheenoh:
> Does the new compiler directory have the patched version of the 2.7 compiler?

This is probably the cause. I forgot the CI check here doesn't run the setup command to do the patch. I'll put it on manually.

Pheenoh (approved):
There's another bug in the CD.yml I noticed this morning. I'll do a follow up PR where I fix that + switch this process to use `./tp setup` going forward.

Will probably add wibo to this process as well to speed up the time it takes to run the check.

PR #381, opened 2023-07-21
Importing more JKernel code

hatal175:


TakaRikka (approved):

include/dolphin/types.h
```diff
@@ -29,6 +29,9 @@ typedef int BOOL;
 #define TRUE 1
 #define FALSE 0
 
+#define READU32_BE(ptr, offset) \
+    (((u32)ptr[offset] << 24) | ((u32)ptr[offset + 1] << 16) | ((u32)ptr[offset + 2] << 8) | (u32)ptr[offset + 3]);
+
```
TakaRikka, line 34: as i think about this more, should there be macros for reading other types from a pointer as well? a lot of the void* accessing in d_stage and other places seems really bad. wonder if they had more macros to work with raw data better

PR #382, opened 2023-07-22
d_menu_skill mostly matching

Trueffeloot:
Last nonmatching is the virtual function draw at the end of the TU. There is a padding problem with the vtable which I didn't know how to fix


hatal175:
Looks good. will merge once conflict is solved.

Trueffeloot:
conflict is fixed

hatal175 (approved):

PR #391, opened 2023-07-26
Izumi_gate+vground

Trueffeloot:
## CC0 License Agreement

- [X] I agree to the terms of the CC0 License.


.github/pull_request_template.md
```diff
@@ -9,7 +9,7 @@ To the best of my knowledge and belief, my contribution is either originally cre
 I understand that this project and its maintainers are not responsible for enforcing the CC0 license, and I release them from any potential liability related to my contribution.
 -->
 
-- [ ] I agree to the terms of the CC0 License.
+- [X] I agree to the terms of the CC0 License.
```
TakaRikka, line 12: this doesn't need to be changed in the file itself, it's only used within github. you can revert the change here

rel/d/a/obj/d_a_obj_vground/d_a_obj_vground.cpp
```diff
@@ -1,357 +1,163 @@
...
+            return 0;
+        } else {
+            J3DAnmTevRegKey* mpTevReqKey = (J3DAnmTevRegKey*)dComIfG_getObjectRes(l_arcName, 8);
+            mpBrk = new mDoExt_brkAnm();
+            if ((!mpBrk) || (!mpBrk->init(mpObjectRes, mpTevReqKey, 1, NULL, 1.0f, NULL, -1))) {
+                return 0;
+            }
+            return 1;
+        }
+    }
```
TakaRikka, line 61: readability can be improved by removing unnecessary nesting. for example, an else statement if not necessary after the returns. you can just do it like

```c++
if (!mpModel) {
    return 0;
}

J3DAnmTextureSRTKey* mpTextureSRTKey = (J3DAnmTextureSRTKey*)dComIfG_getObjectRes(l_arcName, 0xb);
```
and so on so forth for the others.

Trueffeloot:
Function should look better now 👍 I hope it's good like this

Trueffeloot:
Oh yeah and I reverted the .md file

TakaRikka (approved):

PR #1878, opened 2023-08-13
d_a_tag_bottle_item OK, d_a_tag_chgrestart work

Pheenoh:
d_a_tag_chgrestart basically done as well, just couldn't figure out one do-while loop

---

## CC0 License Agreement

- [x] I agree to the terms of the CC0 License.


include/rel/d/a/tag/d_a_tag_chgrestart/d_a_tag_chgrestart.h
```diff
@@ -1,6 +1,26 @@
 #ifndef D_A_TAG_CHGRESTART_H
 #define D_A_TAG_CHGRESTART_H
 
-#include "dolphin/types.h"
+#include "d/com/d_com_inf_game.h"
+#include "f_op/f_op_actor_mng.h"
+#include "JSystem/JKernel/JKRHeap.h"
+#include "dol2asm.h"
```
TakaRikka, line 7: the unnecessary inclusions for the class def here might be better to move to the cpp file instead so that there's less rebuilding to do with changes (also dol2asm.h can probably just be removed entirely if the file doesnt need any of the defines)

rel/d/a/tag/d_a_tag_bottle_item/d_a_tag_bottle_item.cpp
```diff
@@ -1,361 +1,235 @@
...
+    s32 uVar2 = 1;
+  
+    if (!i_dComIfGp_getEvent().i_isOrderOK()) {
+        uVar2 = 0;
+        
+        if (mEvtInfo.i_checkCommandCatch()) {
+            return uVar2;
+        }
+    }
+    return uVar2;
```
TakaRikka, line 130: this is probably a one liner like
```cpp
return i_dComIfGp_getEvent().i_isOrderOK() && mEvtInfo.i_checkCommandCatch();
```
or something like that

rel/d/a/tag/d_a_tag_chgrestart/d_a_tag_chgrestart.cpp
```diff
@@ -1,214 +1,158 @@
...
+            stage_actor_data_class* actor_data = actor->mEntries;
+            int entryNum = actor->mEntryNum;
+
+            do {
+                actor_data++;
+            }
+            while (entryNum > 0 && actor->mEntryNum > playerNo && actor_data != 0);
+            
+            l_restartPos = actor_data->mSpawnPos;
+            l_restartAngle = actor_data->mAngle.y;
```
TakaRikka, line 91: im guessing this whole section is supposed to be within a for loop

rel/d/a/tag/d_a_tag_bottle_item/d_a_tag_bottle_item.cpp
Pheenoh: I can't seem to get a match with a one liner here

TakaRikka (approved):

PR #1920, opened 2023-09-09
Trying to fix author in ok check

hatal175:
I think this should fix this.
Mostly based on https://superuser.com/questions/360966/how-do-i-use-a-bash-variable-string-containing-quotes-in-a-command

I'd appreciate a look though.

## CC0 License Agreement

- [x] I agree to the terms of the CC0 License.


Pheenoh:
The real issue here is looking up the user by display names and not by their actual github username (which can't have spaces). We can merge and see if this fixes though, if not i have a solution that will use the actual github username

Pheenoh (approved):

PR #1954, opened 2023-10-03
d_a_tag_myna_light OK

### LINKED ISSUE #1170: d_a_tag_myna_light.cpp

jakepatzer:
Resolves #1170


include/rel/d/a/tag/d_a_tag_myna_light/d_a_tag_myna_light.h
```diff
@@ -2,5 +2,33 @@
...
+    /* 80D5D148 */ u32 getTurnOnTime();
+    /* 80D5D154 */ u32 getTurnOffTime();
+    /* 80D5D160 */ void initialize();
+
+    /* 80D5D298 */ virtual ~daTag_MynaLight_c();
+
+    inline bool checkTurnOnOffChange() {
+        bool var1 = mTurnOnFlag - field_0x578;
+        return var1 & 0xFF;
+    }
```
TakaRikka, line 23: for inlines defined within the class, you can remove the inline keyword

rel/d/a/tag/d_a_tag_myna_light/d_a_tag_myna_light.cpp
```diff
@@ -1,401 +1,198 @@
...
+    static const u8 color_data[4] = {0xBC, 0x66, 0x42, 0xFF};
+
+    if (mTurnOnFlag) {
+        cLib_addCalc(&field_0x56c, 1.0f, 0.5f, 0.1f, 0.0001f);
+    } else {
+        cLib_addCalc(&field_0x56c, 0.0f, 0.5f, 0.1f, 0.0001f);
+    }
+
+    if (field_0x56c >= 0.000001f) {
+        *(u32*)&color = *(u32*)color_data;
```
TakaRikka, line 72: can this function be changed to initialize the GXColor here like `GXColor color = {0xBC, 0x66, 0x42, 0xFF}`? or does it not match like that

rel/d/a/tag/d_a_tag_myna_light/d_a_tag_myna_light.cpp
```diff
@@ -1,401 +1,198 @@
...
 
 /* 80D5D210-80D5D230 000750 0020+00 1/0 0/0 0/0 .text            daTag_MynaLight_Create__FPv */
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-static asm void daTag_MynaLight_Create(void* param_0) {
-    nofralloc
-#include "asm/rel/d/a/tag/d_a_tag_myna_light/d_a_tag_myna_light/daTag_MynaLight_Create__FPv.s"
+static void daTag_MynaLight_Create(daTag_MynaLight_c* i_this) {
+    i_this->create();
```
TakaRikka, line 148: this applies to all actor methods (Create, Execute, Draw, Delete, IsDelete), they should all have an int return type and usually return the value of the function its calling. so like here it'd be like
```cpp
static int daTag_MynaLight_Create(void* i_this) {
    return static_cast<daTag_MynaLight_c*>(i_this)->create();
}
```

TakaRikka, line 148: i notice you changed the parameter types in the method functions, but the name mangling from the symbol map confirms that void* was what was used originally. the symbol map should be followed as closely as possible where possible. in these cases, you can just use a cast to the correct actor type, like
`static_cast<ActorType*>(i_this)->` or `((ActorType*)i_this)->`

rel/d/a/tag/d_a_tag_myna_light/d_a_tag_myna_light.cpp
jakepatzer: Huh, I swear I tried this before and it didn't work. Must have been doing something wrong though as it's working now, looks much cleaner!

TakaRikka (approved):

PR #1955, opened 2023-10-06
JUTProcBar, JUTResFont

Jcw87:


Jcw87:
JUTGamePad doesn't like this, would have been nice if `make` decided to rebuild the whole library automatically when you change the flags instead of allowing a stale object to give a false OK

PR #1963, opened 2023-10-19
Fix cLib_calcTimer

jakepatzer:


include/SSystem/SComponent/c_lib.h
```diff
@@ -87,7 +87,7 @@ inline T cLib_maxLimit(T val, T max) {
 
 template <typename T>
 T cLib_calcTimer(T* value) {
-    if (*value != 0) {
+    if (*value != (u16)0) {
```
TakaRikka, line 90: are you sure this is right? since it's a template, maybe at least a cast to `T`?

jakepatzer: I'm not 100% sure, but can confirm that casting to T doesn't work (in this case, `T` is` s16` but the `0` must be `u16` for it to match. All other data types, whether signed or unsigned, are non-matching). At the very least, there's a few other existing implementations of this function (of types `s32` and `u8`), neither of which compile any differently as a result of this change. I'm a bit confused as to why that is though

TakaRikka: i guess this is good enough for now from what i can tell, but a comment in the function explaining it might be a fake match might be good incase a cleaner solution is found.

jakepatzer: Good call, making a revision now

TakaRikka (approved):

PR #1997, opened 2023-11-30
d_a_obj_barDesk OK

randomsalience:


rel/d/a/obj/d_a_obj_barDesk/d_a_obj_barDesk.cpp
```diff
@@ -4,705 +4,217 @@
...
-/* ############################################################################################## */
-/* 80BA9D18-80BA9D24 000008 000C+00 1/1 0/0 0/0 .bss             @3643 */
-static u8 lit_3643[12];
-
-/* 80BA9D24-80BA9D2C 000014 0008+00 2/2 0/0 0/0 .bss             l_HIO */
-static u8 l_HIO[8];
-
-/* 80BA9D2C-80BA9D30 00001C 0004+00 1/1 0/0 0/0 .bss             None */
-static u8 data_80BA9D2C[4];
+typedef void (daBarDesk_c::*daBarDesk_modeFunc)();
```
Trueffeloot, line 102: I'd actually move that into the windowProc function since there's no other spot where this typedef is needed

rel/d/a/obj/d_a_obj_barDesk/d_a_obj_barDesk.cpp
```diff
@@ -4,705 +4,217 @@
 //
```
Trueffeloot, line 4: you could create a new nice description to replace the dol2asm TU description. Something like:
image
I have to show it as a screenshot because markdown creates bullet points

Trueffeloot (approved):
lgtm! the comments aren't that important just very small notes so I'll approve anyways

rel/d/a/obj/d_a_obj_barDesk/d_a_obj_barDesk.cpp
randomsalience: Yeah that makes sense

PR #2007, opened 2023-12-19
d_a_obj_bk_leaf OK

### LINKED ISSUE #809: d_a_obj_bk_leaf.cpp

jakepatzer:
Resolves #809 

Could use a sanity check on the `make_prm` and `daBkLeaf_c::setBokkuri` functions. It seems incredibly unlikely these would actually match the originals, especially where we intentionally call a member function on a null pointer. But `make_prm` matches the function signature found in the debug rom as far as I can tell, so for the time being I worked within that constraint for the sake of accuracy.


rel/d/a/obj/d_a_obj_bk_leaf/d_a_obj_bk_leaf.cpp
```diff
@@ -1,363 +1,207 @@
...
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-asm void daBkLeaf_c::setBokkuri() {
-    nofralloc
-#include "asm/rel/d/a/obj/d_a_obj_bk_leaf/d_a_obj_bk_leaf/setBokkuri__10daBkLeaf_cFv.s"
+// fake match?
+void daBkLeaf_c::setBokkuri() {
+    csXyz currentAngle = current.angle;
+    ((daObjCarry_c*)NULL)->make_prm((u32*)&currentAngle, (csXyz*)6, getItem(), getItemBit(), 1, 2);
```
hatal175, line 20: make_prm is probably a static function of daObjCarry_c that returns csXyz.

jakepatzer: It seems I'm unable to get it matching when returning either `csXyz` or `csXyz*`

However it does still match when doing either of the following:
1. Making the function static
2. Changing the parameter types to be "correct":
```
make_prm(u32* param_0, csXyz* param_1, u8 param_2, u8 param_3, u8 param_4, u8 param_5, u8 param_6=0)
->
make_prm(csXyz* param_0, u32 param_1, u8 param_2, u8 param_3, u8 param_4, u8 param_5, u8 param_6=0)
```

In regards to accuracy, how do we tell when Ghidra may be making poor assumptions vs when it is guaranteed to match what's in the rom? for context, this is how this function appears in the debug rom, is it safe to assume that `__thiscall`, as well as the parameter types, may be inaccurate?

```
/* __thiscall daObjCarry_c::make_prm(unsigned long *,
                                     csXyz *,
                                     unsigned char,
                                     unsigned char,
                                     unsigned char,
                                     unsigned char,
                                     unsigned char) */

void __thiscall
daObjCarry_c::make_prm
          (daObjCarry_c *this,ulong *param_1,csXyz *param_2,uchar param_3,uchar param_4,
          uchar param_5,uchar param_6,uchar param_7)
```

hatal175: This works with keeping the parameter types. I do think I was wrong with the return type, it would have set y then - my bad.

static void make_prm(csXyz* param_1, u8 param_2, u8 param_3, u8 param_4, u8 param_5, u8 param_6=0) {
        param_1->x = (param_4 << 8) | param_3;
        param_1->z = param_5 | (param_6 << 13) | ((u32(param_2) & 0xFF) << 1);
    }

void daBkLeaf_c::setBokkuri() {
    csXyz currentAngle = current.angle;
    daObjCarry_c::make_prm(&currentAngle, 6, getItem(), getItemBit(), 1, 2);
    field_0x578 = fopAcM_createChild(PROC_Obj_Carry, fopAcM_GetID(this), 0, &current.pos,
                                     fopAcM_GetRoomNo(this), &currentAngle, &mScale, -1, 0);
}

This worked for me. I don't think ghidra is the best with make stdcall/thiscall decisions.

hatal175 (approved):

PR #2009, opened 2023-12-20
d_meter_HIO OK + misc

hatal175:


Trueffeloot:
lol never noticed that I declared some of the J2DTextbox pointers as [1][6] arrays in dMenuFishing_c! Thanks for fixing that

TakaRikka (approved):

PR #2010, opened 2023-12-22
d_a_obj_chandelier OK

randomsalience:


rel/d/a/obj/d_a_obj_chandelier/d_a_obj_chandelier.cpp
```diff
@@ -1,727 +1,321 @@
...
-#include "asm/rel/d/a/obj/d_a_obj_chandelier/d_a_obj_chandelier/exeModeStartSwg__17daObjChandelier_cFv.s"
+void daObjChandelier_c::exeModeStartSwg() {
+    moveSwing(20.0f, 300.0f, 50.0f, 0.3f);
+    if (mTimer >= 90) {
+        mMode = MODE_DOWN;
+        field_0x5fc = 0;
+        mTimer = 0;
+        shape_angle.y = orig.angle.y;
+        shape_angle.z = orig.angle.z;
+        current.pos = field_0x5a8;
```
Trueffeloot, line 126: Does the debug rom show something here about an inline maybe `fopAcM_SetPos`?

Trueffeloot (approved):
Lgtm! Just one small question but I'll approve it already

rel/d/a/obj/d_a_obj_chandelier/d_a_obj_chandelier.cpp
TakaRikka: nah this is correct as is

TakaRikka (approved):

PR #2019, opened 2024-01-01
libstage.py

jdflyer:
This is a script that can take in any dzs/dzr file from the game, convert the data into a json file, and recreate the original dzs/dzr. I also modified the extract scripts to automatically convert all dzs/dzr files to json, and to automatically convert them back into matching dzs/dzr files during asset packaging.


TakaRikka (approved):
i wonder if we should have a "ZAPD" like tool for tww/tp assets, might be interesting. lgtm for now though

PR #2042, opened 2024-01-22
Fix _stack_addr to match OSInit and __OSThreadInit

hatal175:


tools/lcf.py
```diff
@@ -203,7 +203,7 @@ def lcf_generate(output_path,shiftable,map_file):
 
         file.write("\t} > text\n")
         file.write(
-            "\t_stack_addr = (_f_sbss2 + SIZEOF(.sbss2) + 65536 + 0x7) & ~0x7;\n"
+            "\t_stack_addr = (_f_sbss2 + SIZEOF(.sbss2) + 0x1000 + 0x7) & ~0x7;\n"
```
TakaRikka, line 206: is this a hack or is it likely that this was originally done? just curious cause this is the first instance i've seen of the lcf being different among GC decomp projects

hatal175: https://github.com/PrimeDecomp/prime/blob/main/ldscript.lcf is different. And the fix is basically from 0x10000 to 0x1000.

encounter: Yeah, stack size was configurable in the CW IDE. Though I've only ever seen _increases_, not decreases 🤔

TakaRikka: i'll assume this is correct for now then, unless it causes issues later on

TakaRikka (approved):

PR #2066, opened 2024-02-06
d_a_obj_wind_stone OK

Trueffeloot:


hatal175 (approved):

hatal175:
Just need to solve conflict

Trueffeloot:
Done! About diff.py file change I just chmod +x'ed it locally and accidentally committed and pushed it too but I think this isn't an issue

PR #2074, opened 2024-02-17
d_map OK

TakaRikka:
`dDlst_base_c::draw` is supposed to be autogenerated at the of d_map, but was having issues with it. it should fix other issues with inheriting dDlst_base_c though if it can be generated correctly, so need to look into again at some point


Trueffeloot:
I'm 100% sure that dDlst_c::draw needs to be defined instead of declared in the header. But that changes so much vtable padding that would need to be fixed and the whole build doesn't match this way currently (even the rels). I believe we have to decompile more or even all classes that inherit from dDlst_base_c and decompile some of their functions (specifically ctors and dtors and virtuals which use the vtable) in order to match the vtable padding. So this will probably take a while if I'm right about what I said

jdflyer:
@Trueffeloot dDlst_base_c::draw is generated. I was looking at the movie player REL a while back and the dDlst_base_c::draw function also gets generated there

Trueffeloot:
> @Trueffeloot dDlst_base_c::draw is generated. I was looking at the movie player REL a while back and the dDlst_base_c::draw function also gets generated there

Yeah it's actually generated in multiple rels for some reason... I don't really understand why

jdflyer (approved):

PR #2076, opened 2024-02-25
d_a_obj_table OK

GinNoOokami:


TakaRikka:
project setup seems to have broke for some reason, not your fault. will have to look into it

hatal175 (approved):

include/rel/d/a/obj/d_a_obj_table/d_a_obj_table.h
```diff
@@ -2,8 +2,29 @@
...
 
+class daObjTable_c : public dBgS_MoveBgActor {
+public:
+    /* 80D06560 */ int CreateHeap();
+    /* 80D065D0 */ int Create();
+    /* 80D0695C */ int Execute(Mtx**);
+    /* 80D06AD8 */ int Draw();
+    /* 80D06BC8 */ int Delete();
+
+    request_of_phase_process_class& getPhase() { return mPhaseReq; }
```
TakaRikka, line 16: since this isnt a real function (based on the debug rom), this can be removed. most likely the original didn't have the members private, so you can just make all of the members here public and remove the fake getter

include/JSystem/J3DGraphAnimator/J3DModel.h
```diff
@@ -95,6 +95,8 @@ class J3DModel {
     MtxP i_getAnmMtx(int p1) { return mMtxBuffer->getAnmMtx(p1); }
     void i_setAnmMtx(int p1, Mtx mtx) { mMtxBuffer->setAnmMtx(p1, mtx); }
 
+    inline J3DMaterial* getMaterialNode(u32 index) { return getModelData()->getMaterialTable().getMaterialNodePointer(index); }
```
TakaRikka, line 98: this is a fake function so it can be removed. the correct function to use instead would be `getMaterialNodePointer` which we already have setup

include/rel/d/a/obj/d_a_obj_table/d_a_obj_table.h
```diff
@@ -2,8 +2,29 @@
...
+
+private:
+    inline s16 getFlowID() { return (s16)fopAcM_GetParamBit(this, 0, 16); }
```
TakaRikka, line 19: this also appears to be a fake function? i think what you want here is `getMessageNo` based on debug rom output, which would look like
```c++
u16 getMessageNo() { return fopAcM_GetParam(this); }
```

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
```diff
@@ -116,129 +32,176 @@ static int daObjTable_IsDelete(daObjTable_c* i_this) {
...
 /* 80D064B0-80D06560 0000F0 00B0+00 1/0 0/0 0/0 .text            daObjTable_Create__FP10fopAc_ac_c
  */
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-static asm int daObjTable_Create(fopAc_ac_c* i_this) {
-    nofralloc
-#include "asm/rel/d/a/obj/d_a_obj_table/d_a_obj_table/daObjTable_Create__FP10fopAc_ac_c.s"
+static void daObjTable_Create(fopAc_ac_c* i_this) {
+    daObjTable_c* pTable = static_cast<daObjTable_c*>(i_this);
```
TakaRikka, line 43: personal preference, but i like to use `a_this` for the casted pointer since it's more obvious it's a c-style `this`, and there's some evidence the original devs also used it

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
```diff
@@ -116,129 +32,176 @@ static int daObjTable_IsDelete(daObjTable_c* i_this) {
...
 
 /* 80D064B0-80D06560 0000F0 00B0+00 1/0 0/0 0/0 .text            daObjTable_Create__FP10fopAc_ac_c
  */
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-static asm int daObjTable_Create(fopAc_ac_c* i_this) {
-    nofralloc
-#include "asm/rel/d/a/obj/d_a_obj_table/d_a_obj_table/daObjTable_Create__FP10fopAc_ac_c.s"
+static void daObjTable_Create(fopAc_ac_c* i_this) {
```
TakaRikka, line 42: this function, like all actor method functions, should be returning an int. you can probably fix this by doing something like
```c++
int phase_state = dComIfG_resLoad(a_this->mPhase), l_arcName);
if (phase_state == cPhs_COMPLEATE_e) {
    if (a_this->MoveBGCreate(l_arcName, 8, NULL, 0x4000, NULL) == cPhs_ERROR_e) {
        return cPhs_ERROR_e;
    }
}

return phase_state;
```

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
```diff
@@ -116,129 +32,176 @@ static int daObjTable_IsDelete(daObjTable_c* i_this) {
...
+        mAttentionInfo.mFlags = 0x2000000a;
+
+        mAttentionInfo.field_0x0[0] = dKy_darkworld_check() ? 218 : 98;
+        mAttentionInfo.field_0x0[1] = mAttentionInfo.field_0x0[0];
+        mAttentionInfo.field_0x0[3] = mAttentionInfo.field_0x0[0];
+    }
+
+    f32 z = current.pos.z;
+    f32 y = current.pos.y + 100.0f;
+    mAttentionInfo.mPosition.x = current.pos.x;
```
TakaRikka, line 110: this can probably be simplified into
```c++
mAttentionInfo.mPosition.set(current.pos.x, current.pos.y + 100.0f, current.pos.z);
```

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
```diff
@@ -116,129 +32,176 @@ static int daObjTable_IsDelete(daObjTable_c* i_this) {
...
+    mMsgFlow.init(this, getFlowID(), 0, NULL);
+    dMsgObject_endFlowGroup();
+
+    mCullMtx = mpModel->mBaseTransformMtx;
+
+    mDoMtx_stack_c::transS(current.pos);
+    mDoMtx_YrotM(mDoMtx_stack_c::get(), shape_angle.y);
+
+    PSMTXCopy(mDoMtx_stack_c::get(), mBgMtx);
+    PSMTXCopy(mDoMtx_stack_c::get(), mpModel->mBaseTransformMtx);
```
TakaRikka, line 123: interesting thing going on here, this appears to be a func that got inlined. usually they don't, but in this case the debug rom reveals it

so what you can do is in the class body declare it as an explicit inline like
```c++
inline void initBaseMtx();
inline void setBaseMtx();
```

and then above this function in the source file, define the new functions similarly
```c++
void daObjTable_c::initBaseMtx() {
    fopAcM_SetMtx(this, mpModel->i_getBaseTRMtx());
    setBaseMtx();
}

void daObjTable_c::setBaseMtx() {
    mDoMtx_stack_c::transS(current.pos);
    mDoMtx_stack_c::YrotM(shape_angle.y);
    cMtx_copy(mDoMtx_stack_c::get(), mBgMtx);
    mpModel->setBaseTRMtx(mDoMtx_stack_c::get());
}
```

so then finally in the create function, you can correct it into
```c++
initBaseMtx();
mEventID = -1;
return cPhs_COMPLEATE_e;
```

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
```diff
@@ -116,129 +32,176 @@ static int daObjTable_IsDelete(daObjTable_c* i_this) {
...
+        } else {
+            mpModel->getMaterialNode(1)->getShape()->onFlag(J3DShpFlag_Visible);
+        }
+    }
+
+    if (i_dComIfGs_isEventBit(dSv_event_flag_c::saveBitLabels[103]) &&
+        !(!i_dComIfGs_isEventBit(dSv_event_flag_c::saveBitLabels[268]) &&
+          i_dComIfGs_isEventBit(dSv_event_flag_c::saveBitLabels[264])))
+    {
+        mpModel->getMaterialNode(4)->getShape()->onFlag(J3DShpFlag_Visible);
```
TakaRikka, line 98: all of these getShape()->onFlag calls can be reduced to `getShape()->hide()`, based on debug rom

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
```diff
@@ -116,129 +32,176 @@ static int daObjTable_IsDelete(daObjTable_c* i_this) {
...
+                mEventID = i_dComIfGp_getEventManager().getEventIdx(this, "TABLE_MAP", -1);
+
+                fopAcM_orderChangeEventId(this, mEventID, 1, -1);
+            }
+        } else if (i_dComIfGp_getEventManager().getMyStaffId(l_arcName, NULL, 0) != -1 &&
+                   mEvtInfo.i_checkCommandDemoAccrpt() && mEventID != -1 &&
+                   i_dComIfGp_getEventManager().endCheck(mEventID))
+        {
+            mEventID = -1;
+            g_meter2_info.setPauseStatus(7);
```
TakaRikka, line 151: this should be `dMeter2Info_setPauseStatus`

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
```diff
@@ -116,129 +32,176 @@ static int daObjTable_IsDelete(daObjTable_c* i_this) {
...
-#pragma optimization_level 0
-#pragma optimizewithasm off
-asm int daObjTable_c::Execute(Mtx** i_mtx) {
-    nofralloc
-#include "asm/rel/d/a/obj/d_a_obj_table/d_a_obj_table/Execute__12daObjTable_cFPPA3_A4_f.s"
+int daObjTable_c::Execute(Mtx** i_mtx) {
+    mEvtInfo.i_onCondition(1);
+
+    dComIfG_inf_c& gameInfo = g_dComIfG_gameInfo;
+    if (gameInfo.getPlay().getEvent().runCheck()) {
```
TakaRikka, line 136: i assume you already tried to use inlines such as `i_dComIfGp_event_runCheck` and `dComIfGp_getEvent().reset()` and had matching issues, so if thats the case then put a comment here just noting its a bit of a fake match

TakaRikka (commented):
looks pretty good, just a few minor things to note that aren't as obvious starting out. they mostly revolve around using the debug rom as a greater reference

include/JSystem/J3DGraphAnimator/J3DModel.h
GinNoOokami: A general question I have about this. Since the debug rom (presumably) doesn't have anything stripped, is the goal to avoid introducing any methods that aren't present, even if they lead to greater readability or clarity?

TakaRikka: yep, we're trying to follow the original dev implementation as close as possible with what publicly available info we have. any sort of modern refactoring would be better left to a separate project with that goal in mind

GinNoOokami: Understood.

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
GinNoOokami: That makes sense, I didn't notice the set method.

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
GinNoOokami: Thanks for explaining this. Your suggestion works, except I had to replace the last line with `i_setBaseTRMtx` because the retail rom expects the `MTXCopy`.

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
GinNoOokami: I was able to still get a match with your first suggestion, but I couldn't find any other combination that still matches for the event `reset` method.

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
GinNoOokami: It makes sense that all actors _should_ have a consistent signature, but I was unable to match by returning any sort of value. Naturally, I started out with something similar to your example, but no matter what variations I tried there were some stack and branch differences that didn't line up with the retail.

The problem seems to be that the return values from both calls are placed directly in the r3 register. Placing it in a stack variable generates unwanted `mv` instructions, while returning enum values generates unwanted `li` instructions. I can't seem to find an arrangement of branching in the high level code that compiles to match the efficient asm code.

GinNoOokami:
I think I've addressed all current feedback with the exception of the possible fake match for the `event::reset()` call and the return value for the static create. Open to suggestions for either of those two.

rel/d/a/obj/d_a_obj_table/d_a_obj_table.cpp
TakaRikka: if you have a decomp.me scratch for it i can take a look, but if not its alright i can fix another time

GinNoOokami: https://decomp.me/scratch/9nBrK

GinNoOokami: The method ends at 0xac in that scratch. There's some runoff into another method, but I'm not sure how to fix that.

TakaRikka (approved):

PR #2089, opened 2024-03-09
d_a_obj_rock small cleanup and doc pass

GinNoOokami:


include/rel/d/a/obj/d_a_obj_rock/d_a_obj_rock.h
```diff
@@ -26,14 +26,14 @@ STATIC_ASSERT(sizeof(rock_ss) == 0x160);
 class obj_rock_class : public fopAc_ac_c {
 public:
     /* 0x568  */ request_of_phase_process_class mPhaseReq;
-    /* 0x570  */ u8 field_0x570;
-    /* 0x571  */ u8 field_0x571;
-    /* 0x572  */ u8 field_0x572;
-    /* 0x574  */ s16 field_0x574;
+    /* 0x570  */ u8 mDesiredRockNum;
```
TakaRikka, line 29: im guessing this class acts as a handler for a patch of rocks based on usage, so i'd probably change some of these to like
```
mDesiredRockNum -> mPatchRockNum
mRocksInitialized -> mPatchInitialized
mCount -> mRockNum
```

rel/d/a/obj/d_a_obj_rock/d_a_obj_rock.cpp
```diff
@@ -120,15 +120,16 @@ static int daObj_Rock_Create(fopAc_ac_c* i_this) {
...
 
         if (a_this->field_0x572 == 0xFF) {
             a_this->field_0x572 = 0;
         }
 
-        a_this->mCount = a_this->field_0x570 + 1;
+        // Unclear why a separate counter is used here, or why it adds 1
+        a_this->mCount = a_this->mDesiredRockNum + 1;
         if (a_this->mCount > 0x10) {
             a_this->mCount = 0x10;
```
TakaRikka, line 134: in cases where it's reasonable to assume decimal was used, like a counter, it probably makes more sense to use rather than hex. so i'd use 16 here over 0x10

TakaRikka (approved):

PR #2100, opened 2024-03-19
Match most of d_file_select

hatal175:


Trueffeloot:
oh no i was also working on that ;( maybe we should actually start assigning if it's work on the main.dol

hatal175:
It was exceedingly tedious if it's any comfort

TakaRikka (approved):

PR #2103, opened 2024-03-20
fix setLightTevColorType_MAJI usage

randomsalience:
This minor fix corrects files that used the wrong version of the overloaded setLightTevColorType and setLightTevColorType_MAJI functions


Trueffeloot (approved):
good catch!

TakaRikka (approved):

PR #2151, opened 2024-05-08
d_a_tag_sppath OK

dooplecks:


rel/d/a/tag/d_a_tag_sppath/d_a_tag_sppath.cpp
```diff
@@ -1,365 +1,231 @@
...
 
 /* 80D62BA0-80D62BC0 001000 0020+00 1/0 0/0 0/0 .text daTagSppath_Execute__FP13daTagSppath_c */
-#pragma push
-#pragma optimization_level 0
-#pragma optimizewithasm off
-static asm void daTagSppath_Execute(daTagSppath_c* param_0) {
-    nofralloc
-#include "asm/rel/d/a/tag/d_a_tag_sppath/d_a_tag_sppath/daTagSppath_Execute__FP13daTagSppath_c.s"
+static void daTagSppath_Execute(daTagSppath_c* tag) {
+    tag->execute();
```
TakaRikka, line 198: should return the execute ret value here as an int

rel/d/a/tag/d_a_tag_sppath/d_a_tag_sppath.cpp
```diff
@@ -1,365 +1,231 @@
...
 }
-#pragma pop
 
 /* 80D62BC0-80D62BC8 001020 0008+00 1/0 0/0 0/0 .text            daTagSppath_Draw__FP13daTagSppath_c
  */
-static bool daTagSppath_Draw(daTagSppath_c* param_0) {
+static bool daTagSppath_Draw(daTagSppath_c* tag) {
     return true;
```
TakaRikka, line 204: should return as int

TakaRikka (approved):
looks good to me. just a couple of non-essential notes but good to know for first time contributors:
- we like to prefix function parameters with `i_` most of the time since it seems to generally be how the original devs did, unless a debug string says otherwise.
- you can use f32 instead of float just for consistency with the other dolphin types
- for the C style actor functions (`daTagSppath_Create`, etc) you can name the actor parameter `i_this`.

TakaRikka:
this should be good to go after fixing progress conflict

PR #2238, opened 2024-11-08
dolphin work

randomsalience:
matched a few functions, and changed some optimization flags to what they were before the switch to dtk


TakaRikka (approved):
looks good. keep in mind that once https://github.com/doldecomp/dolsdk2004 is more complete, i'd like to copy it here as best as possible. have been meaning to do some more work on it so we can copy it over here

PR #2257, opened 2024-12-06
d_a_obj_wood_pendulum done

hev7:


src/d/actor/d_a_obj_wood_pendulum.cpp
```diff
@@ -4,90 +4,25 @@
...
+    mpModel->setBaseScale(scale);
+    setBaseMtx();
 }
 
 /* 80D39454-80D394CC 0000D4 0078+00 2/2 0/0 0/0 .text            setBaseMtx__13daObjWPndlm_cFv */
 void daObjWPndlm_c::setBaseMtx() {
-    // NONMATCHING
+    PSMTXTrans(mDoMtx_stack_c::now, current.pos.x, current.pos.y, current.pos.z);
+    mDoMtx_ZXYrotM(mDoMtx_stack_c::now, shape_angle.x, shape_angle.y, shape_angle.z + field_0x958);
+    MTXCopy(mDoMtx_stack_c::get(), mpModel->mBaseTransformMtx);
```
TakaRikka, line 23: these function calls should be mDoMtx_stack_c::transS, mDoMtx_stack_c::ZXYrotM, mpModel->setBaseTRMtx in that order

src/d/actor/d_a_obj_wood_pendulum.cpp
```diff
@@ -113,177 +46,161 @@ const static dCcD_SrcSph l_sph_src2 = {
...
+    mDoMtx_stack_c::ZXYrotM(shape_angle.x, shape_angle.y, shape_angle.z + field_0x958);
+    mDoMtx_stack_c::multVec(&local_48, &local_48);
+    mSph.SetC(local_48);
+    dComIfG_Ccsp()->Set(&mSph);
+
+    f32 lit_3826[6] = {-60.0f, -1120.0f, 100.0f, 80.0f, -1230.0f, -50.0f};
+    for (int i = 0; i < 2; i++) {
+        local_48.x = lit_3826[i * 3];
+        local_48.y = lit_3826[i * 3 + 1];
+        local_48.z = lit_3826[i * 3 + 2];
```
TakaRikka, line 136: this looks like a 2D array, might be able to simplify the code here?

TakaRikka: actually now that i look at debug closer, this is probably a Vec array and doing something like
```
local_48 = sp38[i];
```

hev7: Thanks, how does it look now?

TakaRikka: 👍

TakaRikka (approved):

PR #2261, opened 2024-12-09
Work on d_a_obj_lv3Water

hev7:
This appears to match in objdiff but fails when building


include/d/d_com_inf_game.h
```diff
@@ -3970,6 +3970,10 @@ inline void dComIfGd_setXluListBG() {
     g_dComIfG_gameInfo.drawlist.setXluListBG();
 }
 
+inline void dComIfGd_setXluListDarkBG() {
```
hev7, line 3973: I see a similar change was made in the other open PR, so I can remove this once the other PR is merged.

hatal175: Merged. You may remove it after rebasing/merging

hev7:
> Regarding the mismatch, the functions are in the wrong order. We have an open issue we don't know how to solve with dEvLib_callback_c inline virtual function generation.

Ah I see, thanks.

hatal175 (approved):

PR #2281, opened 2025-01-03
d_a_obj_waterPillar + d_a_obj_syRock Equivalent (weak ordering)

marwernerh:
* Added fake struct padding (comes before dEvLib_callback_c inheritance)
* Basic documentation
* Added fake Chk inlines to dBgS_Acch
* Added AT_TYPE_100 to cCcD_ObjAtType


include/d/actor/d_a_obj_syRock.h
```diff
@@ -1,57 +1,137 @@
...
+
+    // Switch to determine whether or not the stalactite should be broken automatically when player is in range
+    inline u32 getSwBit2() {
+        return fopAcM_GetParamBit(this, 8, 8);
+    }
+
+    // Event ID for Midna to stop and note the stalactite that fell by itself
+    inline s32 getEvetID() {
+        return fopAcM_GetParamBit(this, 16, 8);
+    }
```
TakaRikka, line 95: not a major thing but just for future reference, functions you define in the class like this don't need the inline keyword. they will automatically be inlined

src/d/actor/d_a_obj_syRock.cpp
```diff
@@ -4,717 +4,467 @@
...
-SECTION_DEAD static char const* const stringBase_80D04030 = "syourock";
-#pragma pop
+    mpModels[mIsUnbroken]->setBaseScale(scale);
 
-/* 80D02454-80D02584 000294 0130+00 1/0 0/0 0/0 .text            CreateHeap__10daSyRock_cFv */
-void daSyRock_c::CreateHeap() {
-    // NONMATCHING
+    // Set base transform matrix of current model and the background matrix both to the contents of now
+    mpModels[mIsUnbroken]->setBaseTRMtx(mDoMtx_stack_c::get());
+    PSMTXCopy(mDoMtx_stack_c::get(), mBgMtx);
```
TakaRikka, line 52: another minor thing, you can leave off the PS prefix and just use MTXCopy for example

TakaRikka (approved):

PR #2287, opened 2025-01-18
d_a_obj_octhashi equivalent except for weak func ordering of cCcD_GStts dtor

YunataSavior:
image


src/d/actor/d_a_obj_octhashi.cpp
```diff
@@ -659,30 +401,42 @@ static bool daObjOCTHASHI_IsDelete(daObjOCTHASHI_c* param_0) {
...
 
 /* 80CA6444-80CA6478 001904 0034+00 1/0 0/0 0/0 .text            Delete__15daObjOCTHASHI_cFv */
-void daObjOCTHASHI_c::Delete() {
-    // NONMATCHING
+int daObjOCTHASHI_c::Delete() {
+    dComIfG_resDelete(&mPhaseReq, l_arcName[0]);
+    return 1;
 }
 
+AUDIO_INSTANCES
```
hatal175, line 440: This is meant for singleton instances of the Z2 classes that appear in some of the files. It is not needed here

src/d/actor/d_a_obj_octhashi.cpp
```diff
@@ -659,30 +401,42 @@ static bool daObjOCTHASHI_IsDelete(daObjOCTHASHI_c* param_0) {
...
-/* 80CA625C-80CA62A4 00171C 0048+00 1/0 0/0 0/0 .text            __dt__10cCcD_GSttsFv */
-// cCcD_GStts::~cCcD_GStts() {
-extern "C" void __dt__10cCcD_GSttsFv() {
-    // NONMATCHING
-}
-
 /* 80CA62A4-80CA62F8 001764 0054+00 1/0 0/0 0/0 .text            Create__15daObjOCTHASHI_cFv */
-void daObjOCTHASHI_c::Create() {
-    // NONMATCHING
+int daObjOCTHASHI_c::Create() {
```
hatal175, line 405: You can try defining these in the header file. I'm not sure if it would work and I haven't seen it before but it's worth a shot to see if if will make !cCcD_GStts appear in the correct place

include/d/actor/d_a_obj_octhashi.h
```diff
@@ -23,15 +24,43 @@ class daObjOCTHASHI_c : public fopAc_ac_c {
...
-    /* 80CA6444 */ void Delete();
+    /* 80CA5AE0 */ int CreateHeap();
+    /* 80CA5B98 */ int create();
+    /* 80CA62A4 */ int Create();
+    /* 80CA62F8 */ int Execute(Mtx**);
+    /* 80CA636C */ int Draw();
+    /* 80CA6444 */ int Delete();
+
+    void Set_0x6e8(s16 val) { field_0x6e8 = val; }
+    Z2Creature* GetSound() { return &mSound; }
```
TakaRikka, line 35: neither of these functions exist in debug, so probably should try to remove them if possible

src/d/actor/d_a_obj_octhashi.cpp
```diff
@@ -152,50 +30,10 @@ const static dCcD_SrcSph ccSphSrc = {
...
-
 /* 80CA6550-80CA6550 0000D0 0000+00 0/0 0/0 0/0 .rodata          @stringBase0 */
-#pragma push
-#pragma force_active on
-SECTION_DEAD static char const* const stringBase_80CA6550 = "Octhashi";
-#pragma pop
-
-/* 80CA6584-80CA6588 -00001 0004+00 3/3 0/0 0/0 .data            l_arcName */
-SECTION_DATA static void* l_arcName = (void*)&d_a_obj_octhashi__stringBase0;
+static const char* l_arcName[1] = {
```
TakaRikka, line 34: since this is in .data section, its probably not a const char* but just char*

src/d/actor/d_a_obj_octhashi.cpp
```diff
@@ -212,313 +50,292 @@ static dCcD_SrcCyl ccCylSrc = {
...
 void daObjOCTHASHI_c::SphAction() {
-    // NONMATCHING
+    // The following replace the unused rodata lits:
+    f32 padding[1] = {0.0f};
+    f32 fvals_3887[2] = {1.75f, 0.0f};
+    f32 fvals_3888[2] = {2.125f, 0.0f};
+    f32 fvals_3889[2] = {0.0f, 0.0f};
+    f32 fval_3890[1] = {2000.0f};
+    f32 fval_3903[1] = {608.0f};
+    f32 fval_3904[1] = {448.0f};
```
TakaRikka, line 196: if these are necessary i'd prefer they be put in a dummy func rather than here, like above this func create a fake
```
static f32 dummyLiterals()
```
to handle this

include/d/actor/d_a_obj_octhashi.h
```diff
@@ -23,15 +24,43 @@ class daObjOCTHASHI_c : public fopAc_ac_c {
...
+    /* 0x6c8 */ u8 field_0x6c8;
+    /* 0x6c9 */ u8 field_0x6c9[0x6cb - 0x6c9];
+    /* 0x6cb */ u8 field_0x6cb;
+    /* 0x6cc */ s16 field_0x6cc;
+    /* 0x6d0 */ f32 field_0x6d0;
+    /* 0x6d4 */ f32 field_0x6d4;
+    /* 0x6d8 */ f32 field_0x6d8;
+    /* 0x6dc */ s16 field_0x6dc;
+    /* 0x6de */ s16 field_0x6de;
+    /* 0x6e0 */ int mCylinders;
```
TakaRikka, line 52: i think `mCylinders` would be better named as `mPieceNum` or at least `mCylNum` so that it isn't confused with an actual collider

src/d/actor/d_a_obj_octhashi.cpp
YunataSavior: How would I go about putting Create's definition in the header file if it requires the callback?

hatal175: hmmm, you're probably right. I retract this recommendation.

YunataSavior: OK, when I move the functions, this happens:
image

src/d/actor/d_a_obj_octhashi.cpp
YunataSavior: Sure, but what's the return value for?

TakaRikka: its usually used to keep the literals from being stripped out. you can search for other examples of dummyLiterals in the codebase to see how its used there

YunataSavior: When I move them to a static func with no return value, the values don't get stripped out. rodata still matches

TakaRikka: thats fine then, whatever works

hatal175 (approved):

PR #2291, opened 2025-01-22
d_a_obj_kwheel00, 01, & d_a_obj_klift00 Equivalent (weak func and/or vtable order)

marwernerh:
* Three rels equivalent
* Basic documentation for the three rels
* Define explicit constructor for dMdl_obj_c (needed for proper array creation in d_a_obj_klift00)
* Modify d_a_obj_bky_rock to account for new explicit constructor of dMdl_obj_c
* Specify weak func and/or vtable orders for rels in config.py


marwernerh:
Recent work on JHostIO / m_Do_hostIO makes the debug build fail with the current state of the RELs -- will fix in another commit.

hatal175 (approved):

PR #2292, opened 2025-01-23
d_a_obj_TvCdlst matching

YunataSavior:
This was a quick file lol.

Next up: d_a_obj_Y_taihou


src/d/actor/d_a_obj_TvCdlst.cpp
```diff
@@ -5,393 +5,265 @@
...
+        return;
+    }
+    field_0x710 = 0.0f;
 }
 
 /* 80B9F1F0-80B9F220 000610 0030+00 2/2 0/0 0/0 .text            setLight__11daTvCdlst_cFv */
 void daTvCdlst_c::setLight() {
-    // NONMATCHING
+    if (mLightType) {
+        dKy_plight_set((LIGHT_INFLUENCE *) &field_0x714);
```
TakaRikka, line 132: any reason to not setup field_0x714 as LIGHT_INFLUENCE? that should be the correct type

YunataSavior: Because I literally did not know about it 🤣.

I will change this field then resubmit.

YunataSavior: I updated the code based on your feedback.

hatal175 (approved):

PR #2296, opened 2025-01-29
Fix false pointer relocations in data breaking shiftability

LagoLunatic:
DTK detects bytes that look like pointers in unlinked TUs as relocations, these have to be marked as `noreloc` for those unlinked TUs to remain unchanged when modding.

When PAL/other versions are added later they'll also need this fixed, you can search in DTK's asm files for this regex to find relocations that probably aren't real:
`\.4byte .*\+0x`


config/GZ2E01/config.yml
```diff
@@ -365,8 +365,8 @@ modules:
     binary: assets/l_J_hana01_c_01DL.bin
     header: assets/l_J_hana01_c_01DL.h
   - symbol: l_mat2DL
-    binary: assets/l_mat2DL.bin
-    header: assets/l_mat2DL.h
+    binary: assets/l_mat2DL__d_a_grass.bin
+    header: assets/l_mat2DL__d_a_grass.h
```
encounter, line 369: Are these mistakes?

LagoLunatic: Nope, `l_matDL` and `l_mat2DL` aren't globally unique symbol names, there are multiple TUs that have them as local statics. DTK's assets aren't categorized by TU name, so I add the TU name manually to the header name to fix conflicts. In TWW there are 22 different `l_matDL`s, in TP there are 3.

hatal175 (approved):

PR #2297, opened 2025-01-30
ktOnFire setLight()

clarkWakeland:
Decompile of ktOnFire setLight() function. 
First PR for reverse engineering and this repo, wanted to do something very simple to start.

### DIFF
--- a/include/d/actor/d_a_obj_ktOnFire.h
+++ b/include/d/actor/d_a_obj_ktOnFire.h
@@ -18,24 +18,32 @@ public:
     /* 8058C960 */ void lightInit();
     /* 8058C9D4 */ void setLight();
     /* 8058C9F8 */ void cutLight();
-    /* 8058CA1C */ void Execute();
-    /* 8058CDF8 */ bool Draw();
-    /* 8058CE00 */ bool Delete();
+    /* 8058CA1C */ int Execute();
+    /* 8058CDF8 */ int Draw();
+    /* 8058CE00 */ int Delete();
 
     static u8 const mCcDObjInfo[48];
     static u8 mCcDCyl[68];
 
 private:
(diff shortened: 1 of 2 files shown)


src/d/actor/d_a_obj_ktOnFire.cpp
```diff
@@ -273,31 +273,32 @@ bool daKtOnFire_c::Draw() {
...
+bool daKtOnFire_c::Delete() {
+    dKy_plight_cut(&mLight);
+    return true;
 }
 
 /* 8058CE28-8058CE48 000908 0020+00 1/0 0/0 0/0 .text            daKtOnFire_Draw__FP12daKtOnFire_c
  */
 static void daKtOnFire_Draw(daKtOnFire_c* param_0) {
-    // NONMATCHING
+    param_0->Draw();
```
hatal175, line 284: The draw, delete and execute static functions usually have an int return value that propagates the inner function return value.

hatal175 (changes requested):

include/d/actor/d_a_obj_ktOnFire.h
```diff
@@ -20,13 +20,14 @@ class daKtOnFire_c : public fopAc_ac_c {
     /* 8058C9F8 */ void cutLight();
     /* 8058CA1C */ void Execute();
     /* 8058CDF8 */ bool Draw();
-    /* 8058CE00 */ void Delete();
+    /* 8058CE00 */ bool Delete();
```
TakaRikka, line 23: just adding on to what hatal said, all actors generally have a create, delete, draw and execute function, and sometimes a createHeap and isDelete function, which should all return int values to stay consistent

hatal175 (approved):

PR #2299, opened 2025-02-05
copy dolsdk2004 to tp / b_bh + e_mb done

TakaRikka:


YunataSavior:
I would note in the PR title that d_a_b_bh is matching.

PR #2311, opened 2025-02-26
d_a_e_ym (a.k.a. Twilight bugs) 99% equivalent (daE_YM_c::checkBeforeBg and daE_YM_c::create have regalloc issues)

YunataSavior:


include/SSystem/SComponent/c_bg_s_lin_chk.h
```diff
@@ -28,7 +28,7 @@ class cBgS_LinChk : public cBgS_Chk, public cBgS_PolyInfo {
     void ct();
     void Set2(const cXyz*, const cXyz*, unsigned int);
     void PreCalc();
-    void GetCross();
+    cXyz* GetCross() { return &mLin.GetEnd(); }
```
TakaRikka, line 31: this should return a reference, and you can remove the i_GetCross fake. be sure you've ensured no regressions by comparing a diff report before this change and after.

src/d/actor/d_a_e_ym.cpp
```diff
@@ -677,68 +305,207 @@ bool daE_YM_c::checkWallCrash() {
...
+    if (field_0x6f6 != 0) {
+        return 0;
+    }
+    int val = field_0x6fc;
+    if (val != 0) {
+        return 0;
+    }
+    if (action != 8) {
+        bool truth = dComIfGp_getAttention().LockonTruth();
+        if (truth) {
```
TakaRikka, line 353: are some of these temp's necessary? they seem like they can be removed at a glance, but check if it affects regalloc

src/d/actor/d_a_e_ym.cpp
```diff
@@ -677,68 +305,207 @@ bool daE_YM_c::checkWallCrash() {
...
+                    setActionMode(ACT_FLY);
+                } else {
+                    setActionMode(ACT_SURPRISE);
+                }
+                return 1;
+            }
+        } else {
+            field_0x6f8 = 0;
+        }
+    }
```
TakaRikka, line 507: this is a minor style thing but i prefer separating unrelated if statement blocks with an empty line to increase readability.
```cpp
if () {

} else {

}
// empty line
if () {
// . . .
```

src/d/actor/d_a_e_ym.cpp
```diff
@@ -1196,37 +2002,299 @@ void daE_YM_c::setInclination() {
...
 /* 8081013C-8081030C 0080FC 01D0+00 1/1 0/0 0/0 .text            checkRailDig__8daE_YM_cFv */
-void daE_YM_c::checkRailDig() {
-    // NONMATCHING
+u8 daE_YM_c::checkRailDig() {
+    daPy_py_c* player = daPy_getPlayerActorClass();
+    cXyz my_vec_0;
+    if (player->checkWolfDig()) {
+        my_vec_0 = player->getLeftHandPos() - current.pos;
+        if (my_vec_0.abs() < 200.0f) {
+            field_0x714 &= 0xffffff7f;
```
TakaRikka, line 2137: these can usually be simplified to something like `&= ~0x80`

src/d/actor/d_a_e_ym.cpp
```diff
@@ -1248,128 +2316,529 @@ void daE_YM_c::checkElectricStart() {
...
-SECTION_RODATA static f32 const lit_8040 = 2000.0f;
-COMPILER_STRIP_GATE(0x80815AC4, &lit_8040);
-#pragma pop
-
 /* 808120B0-808123C4 00A070 0314+00 1/1 0/0 0/0 .text            setRiverAttention__8daE_YM_cFv */
 void daE_YM_c::setRiverAttention() {
-    // NONMATCHING
+    daPy_py_c* player = daPy_getPlayerActorClass();
+    cXyz player_pos = player->current.pos;
+    attention_info.distances[2] = '<';
```
TakaRikka, line 2629: this looks like a ghidra-ism, should be a numerical value rather than a character

src/d/actor/d_a_e_ym.cpp
```diff
@@ -1539,67 +3180,196 @@ static int daE_YM_Delete(daE_YM_c* i_this) {
...
-    /* empty function */
-}
-
-/* 80813E3C-80813E84 00BDFC 0048+00 1/0 0/0 0/0 .text            __dt__12J3DFrameCtrlFv */
-// J3DFrameCtrl::~J3DFrameCtrl() {
-extern "C" void __dt__12J3DFrameCtrlFv() {
-    // NONMATCHING
+    if (mFlyType == 1) {
+        J3DModelData* model_data = (J3DModelData*)dComIfG_getObjectRes("E_TM", 0x11);
+        JUT_ASSERT(model_data != NULL, 0x1094);
```
TakaRikka, line 3185: these asserts are backwards, its line number first and then the assert condition. additionally, the way the condition is written does actually matter for matching debug and so should match the debug rom's string exactly. `modelData != 0`. variable should be renamed accordingly

include/SSystem/SComponent/c_bg_s_lin_chk.h
YunataSavior: OK, I changed it to return a reference, but I won't remove the i_GetCross fake. I'd rather I left that to you as a cleanup step down the line, since that would require editing a ton of files.

src/d/actor/d_a_e_ym.cpp
YunataSavior: Yeah, I put it there because I knew it would create the correct value, and I was going to clean it up but I forgot :D

src/d/actor/d_a_e_ym.cpp
magcius: For reference, `p == 0` most likely shows up because of the double-interpretation nature of macros. In tww, our JUT_ASSERT macros now do a double-evaluation which causes the string to be `"p == 0"` See e.g. https://godbolt.org/z/PvqsG9asb

src/d/actor/d_a_e_ym.cpp
TakaRikka: im not gonna enforce it for now since just getting files done is more important, but later down the line on future cleanup passes i will enforce it or fix it myself since compact is really unreadable imo. makes understanding whats going on much more difficult

src/d/actor/d_a_e_ym.cpp
TakaRikka: i'll fix this myself in a future pr

TakaRikka (approved):

src/d/actor/d_a_e_ym.cpp
YunataSavior: Alright. In future TUs, I will adopt this coding standard

PR #2320, opened 2025-03-09
daObjNAN_c equivalent

hatal175:


include/d/d_cc_uty.h
```diff
@@ -20,9 +21,8 @@ struct dCcU_AtInfo {
     /* 0x00 */ cCcD_Obj* mpCollider;
     /* 0x04 */ fopAc_ac_c* mpActor;
     /* 0x08 */ Z2Creature* mpSound;
-    /* 0x0C */ u8 field_0xc[2];
-    /* 0x0E */ s16 mHitDirection;
-    /* 0x10 */ u8 field_0x10[4];
+    /* 0x0C */ csXyz mHitDirection;
+    /* 0x12 */ u8 field_0x12[2];
```
YunataSavior, line 25: Can't you omit this padding since csXyz is 6 bytes, thus the compiler will need to fill in 2 bytes to fit the next 4-byte field?

TakaRikka (approved):

PR #2329, opened 2025-03-17
d_a_b_gnd ~86% equivalent

carter-ktb21:


TakaRikka:
make sure you pull in changes from main and fix conflicts. in this case let main overwrite your local changes since its dtk-template stuff

src/d/actor/d_a_b_gnd.cpp
```diff
@@ -347,170 +74,141 @@ SECTION_RODATA static f32 const lit_3819 = 40.0f;
...
+// #pragma force_active on
+// SECTION_DATA static u32 lit_1787[1 + 4 /* padding */] = {
+//     0x02000201,
+//     /* padding */
+//     0x40080000,
+//     0x00000000,
+//     0x3FE00000,
+//     0x00000000,
+// };
+// #pragma pop
```
TakaRikka, line 93: for all data and such that you've set up, it'd be better to just remove the commented out parts. it just kinda bloats everything

src/d/actor/d_a_b_gnd.cpp
```diff
@@ -347,170 +74,141 @@ SECTION_RODATA static f32 const lit_3819 = 40.0f;
...
 
 /* 806029F0-806029F4 000020 0004+00 1/1 0/0 0/0 .data            l_color$3983 */
-SECTION_DATA static u8 l_color[4] = {
-    0x14,
-    0x0F,
-    0x00,
-    0xFF,
+static _GXColor l_color = {
+    20, 15, 0, -1
```
TakaRikka, line 97: for any data that ends in `$XXXX`, this means that its an "in-function static" and so can be moved into the body of the function that it's used in. also since color values are all unsigned, it'd be better to use 255 (0xFF) rather than -1

src/d/actor/d_a_b_gnd.cpp
```diff
@@ -521,2584 +219,6499 @@ SECTION_DATA static u8 e_name_6757[4] = {
...
-    (void*)NULL /* RTTI */,
-    (void*)NULL,
-    (void*)__dt__10dCcD_GSttsFv,
-};
+/* 805F4F38-805F53A4 0005D8 046C+00 1/0 0/0 0/0 .text            daB_GND_Draw__FP11b_gnd_class */
+static int daB_GND_Draw(b_gnd_class* i_this) {
+     // NONMATCHING
+    if (i_this->field_0x2738) {
+        return 1;
+    }  else {
```
TakaRikka, line 576: in cases where there's an early return like this with the rest of the block in one scope, i like to remove the else statement since it functionally does nothing anyway. helps clean it up a little

src/d/actor/d_a_b_gnd.cpp
```diff
@@ -521,2584 +219,6499 @@ SECTION_DATA static u8 e_name_6757[4] = {
...
+                    } else {
+                        if (i_this->mpModelMorf->checkFrame(32.0f)) {
+                            i_this->mZ2Creature.startCreatureVoice(Z2SE_EN_GND_ATTACK_G2, -1);
+                            i_this->mZ2Creature.startCreatureVoice(Z2SE_EN_GND_V_ATTACK_G2, -1);
+                        } else {
+                            if (i_this->mpModelMorf->checkFrame(26.0f)) {
+                                i_this->mZ2Creature.startCreatureVoice(Z2SE_EN_GND_MANTEAU, -1);
+                            }
+                        }
+                    }
```
TakaRikka, line 4726: just gonna use here as an example, but in many patterns like this you can clean these up with just else if's instead of continually nested if's. the less nests the better

src/d/actor/d_a_b_gnd.cpp
```diff
@@ -521,2584 +219,6499 @@ SECTION_DATA static u8 e_name_6757[4] = {
...
+            //     break;
+            // case 42:
+            //     if (i_this->mpModelMorf->checkFrame(12.0f)) {
+            //         i_this->mZ2Creature.startCreatureVoice(Z2SE_EN_GND_V_DOWNUP, -1);
+            //     } else {
+            //         if (i_this->mpModelMorf->checkFrame(27.0f)) {
+            //             i_this->mZ2Creature.startCreatureVoice(Z2SE_EN_GND_LAND, -1);
+            //         }
+            //     }
+            //     b
```
TakaRikka, line 5885: remember that isDelete methods always return int

TakaRikka (commented):
since most of the functions are still nonmatching anyway i mostly just have cleanup comments or just some tips to remember in the future. i'll probably end up cleaning more when i do a pass over the TU anyway

src/d/actor/d_a_b_gnd.cpp
carter-ktb21: I usually leave it only because the data sections are still nonmatching. It makes it easier to search for specific data sections at least for me. Is there an easier way to do so that makes it unnecessary to have them in comments?

src/d/actor/d_a_b_gnd.cpp
YunataSavior: I 1000% agree with Taka here

src/d/actor/d_a_b_gnd.cpp
YunataSavior: I 1000% agree with Taka here as well

PR #2331, opened 2025-03-18
d_a_obj_burnbox matching

### LINKED ISSUE #975: d_a_obj_burnbox.cpp

Aszman:
resolves #975


src/d/actor/d_a_obj_burnbox.cpp
```diff
@@ -1,109 +1,46 @@
...
-    // NONMATCHING
+    mpBoxModel->setBaseScale(scale);
+    setBaseMtx();
 }
 
 /* 8046E6D4-8046E738 0000B4 0064+00 2/2 0/0 0/0 .text            setBaseMtx__14daObjBurnBox_cFv */
 void daObjBurnBox_c::setBaseMtx() {
-    // NONMATCHING
+    mDoMtx_stack_c::transS(current.pos);
+    mDoMtx_YrotM(mDoMtx_stack_c::now, shape_angle.y);
```
TakaRikka, line 22: mDoMtx_YrotM should be mDoMtx_stack_c::YrotM as per the debug rom

src/d/actor/d_a_obj_burnbox.cpp
```diff
@@ -116,193 +53,187 @@ const static dCcD_SrcCyl l_cyl_src = {
...
+        particlePos.y += 75.0f;
+        break;
+    }
+
+    int index = 0;
+    while (index < 5) {
+        mpEmitters[index] = dComIfGp_particle_set(particle_id[index], &particlePos, NULL, &particleScale, 0xff,
+                                                  NULL, -1, NULL, NULL, NULL);
+        index++;
+    }
```
TakaRikka, line 89: this can probably just be a for loop

src/d/actor/d_a_obj_burnbox.cpp
```diff
@@ -116,193 +53,187 @@ const static dCcD_SrcCyl l_cyl_src = {
...
 
 /* 8046E8BC-8046E93C 00029C 0080+00 1/0 0/0 0/0 .text            CreateHeap__14daObjBurnBox_cFv */
-void daObjBurnBox_c::CreateHeap() {
-    // NONMATCHING
+int daObjBurnBox_c::CreateHeap() {
+    u8 type = getType();
+    J3DModelData* boxModelData = (J3DModelData*) dComIfG_getObjectRes(l_arcName[type], l_bmd[type]);
+    if (boxModelData == NULL) {
+        JUT_PANIC(0x106, "modelData != 0");
+    }
```
TakaRikka, line 107: this block can just be a JUT_ASSERT with the string quotes removed. it's just a normal assert macro but only does anything when compiling as debug

src/d/actor/d_a_obj_burnbox.cpp
```diff
@@ -116,193 +53,187 @@ const static dCcD_SrcCyl l_cyl_src = {
...
+    return phaseStep;
 }
 
 /* 8046E9EC-8046EABC 0003CC 00D0+00 1/0 0/0 0/0 .text            Execute__14daObjBurnBox_cFPPA3_A4_f
  */
-void daObjBurnBox_c::Execute(f32 (**param_0)[3][4]) {
-    // NONMATCHING
+int daObjBurnBox_c::Execute(Mtx** mtx) {
+    field_0x724++;
+    if (field_0x724 % 3 == 0 && (--field_0x728, field_0x728 == -0x14)) {
```
TakaRikka, line 140: can you just do `--field_0x728 == -20` here? would look a little cleaner

src/d/actor/d_a_obj_burnbox.cpp
```diff
@@ -116,193 +53,187 @@ const static dCcD_SrcCyl l_cyl_src = {
...
-// cM3dGCyl::~cM3dGCyl() {
-extern "C" void __dt__8cM3dGCylFv() {
-    // NONMATCHING
-}
-
-/* 8046ED8C-8046EDD4 00076C 0048+00 1/0 0/0 0/0 .text            __dt__8cM3dGAabFv */
-// cM3dGAab::~cM3dGAab() {
-extern "C" void __dt__8cM3dGAabFv() {
-    // NONMATCHING
+void daObjBurnBox_create1st(daObjBurnBox_c* i_this) {
```
TakaRikka, line 189: note that all actor methods (create, delete, isDelete, execute, draw) always return int. so all of these actor methods should be returning the return value of the inner class method

src/d/actor/d_a_obj_burnbox.cpp
Aszman: Not really, `--field_0x728 == -20` changes the assembly code and function is no longer matched. Similar solutions also does not generate the same output. Below is the diff of your suggestion.

image

TakaRikka: you could try separating the if statements into two separate ones with the decrement in the middle, but if that doesnt work then no big deal its fine for now

TakaRikka (approved):

PR #2334, opened 2025-03-19
d_a_obj_zdoor matching

### LINKED ISSUE #1000: d_a_obj_zdoor.cpp

Ttobas:
resolves #1000 

As it is my first decomp, I'd appreciate if you point out even minor details.

### DIFF
--- a/src/d/actor/d_a_obj_zdoor.cpp
+++ b/src/d/actor/d_a_obj_zdoor.cpp
@@ -5,93 +5,16 @@
 
 #include "d/actor/d_a_obj_zdoor.h"
 #include "dol2asm.h"
-
+#include "d/actor/d_a_player.h"
 
 //
 // Forward References:
 //
 
-extern "C" static void checkPlayerPos__FP9daZdoor_c();
-extern "C" static void
-doorCoHitCallBack__FP10fopAc_ac_cP12dCcD_GObjInfP10fopAc_ac_cP12dCcD_GObjInf();
-extern "C" static void
-doorTgHitCallBack__FP10fopAc_ac_cP12dCcD_GObjInfP10fopAc_ac_cP12dCcD_GObjInf();
-extern "C" void init_cyl__9daZdoor_cFv();
-extern "C" void set_cyl__9daZdoor_cFv();
-extern "C" void setBaseMtx__9daZdoor_cFv();
-extern "C" void Create__9daZdoor_cFv();
-extern "C" void CreateHeap__9daZdoor_cFv();
-extern "C" void create1st__9daZdoor_cFv();
-extern "C" void Execute__9daZdoor_cFPPA3_A4_f();
-extern "C" void Draw__9daZdoor_cFv();
-extern "C" void Delete__9daZdoor_cFv();
 extern "C" static void daZdoor_create1st__FP9daZdoor_c();
-extern "C" void __dt__8cM3dGCylFv();
-extern "C" void __dt__8cM3dGAabFv();
-extern "C" void __dt__8dCcD_CylFv();
-extern "C" void __ct__8dCcD_CylFv();
-extern "C" void __dt__12dBgS_ObjAcchFv();
 extern "C" static void daZdoor_MoveBGDelete__FP9daZdoor_c();
 extern "C" static void daZdoor_MoveBGExecute__FP9daZdoor_c();
 extern "C" static void daZdoor_MoveBGDraw__FP9daZdoor_c();
-extern "C" static void func_80D401D4();
-extern "C" static void func_80D401DC();
-extern "C" extern char const* const d_a_obj_zdoor__stringBase0;
-
-//
-// External References:
-//
-
-extern "C" void mDoMtx_YrotS__FPA4_fs();
-extern "C" void mDoMtx_YrotM__FPA4_fs();
-extern "C" void mDoExt_modelUpdateDL__FP8J3DModel();
-extern "C" void mDoExt_J3DModel__create__FP12J3DModelDataUlUl();
-extern "C" void fopAcM_setCullSizeBox2__FP10fopAc_ac_cP12J3DModelData();
-extern "C" void dComIfG_resLoad__FP30request_of_phase_process_classPCc();
-extern "C" void dComIfG_resDelete__FP30request_of_phase_process_classPCc();
-extern "C" void dComIfGp_getReverb__Fi();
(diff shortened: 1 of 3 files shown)


src/d/actor/d_a_obj_zdoor.cpp
```diff
@@ -5,93 +5,16 @@
...
-extern "C" void Execute__9daZdoor_cFPPA3_A4_f();
-extern "C" void Draw__9daZdoor_cFv();
-extern "C" void Delete__9daZdoor_cFv();
 extern "C" static void daZdoor_create1st__FP9daZdoor_c();
-extern "C" void __dt__8cM3dGCylFv();
-extern "C" void __dt__8cM3dGAabFv();
-extern "C" void __dt__8dCcD_CylFv();
-extern "C" void __ct__8dCcD_CylFv();
-extern "C" void __dt__12dBgS_ObjAcchFv();
 extern "C" static void daZdoor_MoveBGDelete__FP9daZdoor_c();
```
YunataSavior, line 15: You must move METHODS and g_profile to the bottom of the file and delete these externs. You must replace the fake contents of METHODS with the real functions

src/d/actor/d_a_obj_zdoor.cpp
```diff
@@ -174,52 +91,82 @@ static const s16 l_open_speed[3] = {
...
-}
 
-/* ############################################################################################## */
-/* 80D40280-80D40284 000094 0004+00 1/1 0/0 0/0 .rodata          @3679 */
-SECTION_RODATA static f32 const lit_3679 = 3.0f;
-COMPILER_STRIP_GATE(0x80D40280, &lit_3679);
+    daPy_py_c* player = daPy_getPlayerActorClass();
+    cXyz distance_to_player = player->current.pos - i_this->current.pos;
+
+    mDoMtx_stack_c::YrotS(-(i_this->current.angle.y));
```
hatal175, line 103: Remove redundant parenthesis

src/d/actor/d_a_obj_zdoor.cpp
```diff
@@ -174,52 +91,82 @@ static const s16 l_open_speed[3] = {
...
-/* 80D4028C-80D40294 0000A0 0008+00 1/1 0/0 0/0 .rodata          @3714 */
-SECTION_RODATA static u8 const lit_3714[8] = {
-    0x43, 0x30, 0x00, 0x00, 0x80, 0x00, 0x00, 0x00,
-};
-COMPILER_STRIP_GATE(0x80D4028C, &lit_3714);
-
 /* 80D3F570-80D3F65C 0001B0 00EC+00 1/1 0/0 0/0 .text
  * doorTgHitCallBack__FP10fopAc_ac_cP12dCcD_GObjInfP10fopAc_ac_cP12dCcD_GObjInf */
-static void doorTgHitCallBack(fopAc_ac_c* i_this, dCcD_GObjInf* param_1, fopAc_ac_c* param_2,
+static void doorTgHitCallBack(fopAc_ac_c* param_0, dCcD_GObjInf* param_1, fopAc_ac_c* player_fopAc,
```
hatal175, line 139: I don't think this is the player actor class as much as this is the actor of whatever you hit

src/d/actor/d_a_obj_zdoor.cpp
```diff
@@ -255,51 +202,64 @@ static dCcD_SrcCyl l_cyl_src2 = {
...
-#pragma force_active on
-SECTION_RODATA static f32 const lit_3799 = 240.0f;
-COMPILER_STRIP_GATE(0x80D402A0, &lit_3799);
-#pragma pop
-
 /* 80D3F740-80D3F884 000380 0144+00 1/1 0/0 0/0 .text            set_cyl__9daZdoor_cFv */
 void daZdoor_c::set_cyl() {
-    // NONMATCHING
+    cXyz vecs[] = {
+        cXyz(60, 0, 0),
```
hatal175, line 222: This might be just me, but since the cXyz gets floats, it is preferable to explicitly write floats (60.0f for example)

src/d/actor/d_a_obj_zdoor.cpp
```diff
@@ -336,36 +305,92 @@ static char* l_dzbName[3] = {
...
-    // NONMATCHING
+    mDoorType = getType();
+    int phase_state = dComIfG_resLoad(&mPhaseReq, l_arcName[mDoorType]);
+    if (phase_state == cPhs_COMPLEATE_e) {
+        int res_name_index = g_dComIfG_gameInfo.mResControl.getObjectResName2Index(l_arcName[mDoorType], l_dzbName[mDoorType]);
+        
+        u16 estimate_size = estimateSizeTbl[mDoorType];
+        phase_state = MoveBGCreate(l_arcName[mDoorType], res_name_index, 0x0, estimate_size, 0x0);
+
+        if (phase_state == 0x5) {
```
hatal175, line 316: That's cPhs_ERROR_e

src/d/actor/d_a_obj_zdoor.cpp
YunataSavior: I agree with Hatal's advice

src/d/actor/d_a_obj_zdoor.cpp
Ttobas: I'm pretty sure it's the player actor.
The pointer is being compared to dComIfGp_getLinkPlayer() to make sure the hit object is the player actor.

I assume this is done to check from which side of the door the player is standing, which then sets correct direction of the speed when the door will be opened.

src/d/actor/d_a_obj_zdoor.cpp
```diff
@@ -396,105 +421,26 @@ extern actor_process_profile_definition g_profile_Obj_ZDoor = {
...
-extern "C" void __dt__12dBgS_ObjAcchFv() {
-    // NONMATCHING
+    return i_this->create1st();
 }
 
 /* 80D40168-80D40188 000DA8 0020+00 1/0 0/0 0/0 .text            daZdoor_MoveBGDelete__FP9daZdoor_c
  */
 static int daZdoor_MoveBGDelete(daZdoor_c* i_this) {
-    // NONMATCHING
+    i_this->MoveBGDelete();
```
TakaRikka, line 434: make sure all these actor methods actually return the inner function `return i_this->MoveBGDelete();`

hatal175:
In this case, we are looking for the player actor to be the one the
collision was with but it can be other actors as well - that's why we need
the comparison check.

hatal175 (approved):

PR #2341, opened 2025-03-22
d_a_obj_flag 95%

tunamayo04:
d_a_obj_flag almost done
Missing some regalloc issues in one function and some other function at 81% that I give up on


include/d/actor/d_a_obj_flag.h
```diff
@@ -2,11 +2,21 @@
...
 
 class FlagJoint_c {
 public:
     /* 80BEC3BC */ ~FlagJoint_c();
     /* 80BEC4E0 */ FlagJoint_c();
+
+    csXyz joint1;
+    csXyz joint2;
+    csXyz joint3;
+    short rv;
```
TakaRikka, line 19: these should probably still follow the `mName` convention. also we want to consistently use dolphin types like s16 instead of short for game code

include/d/actor/d_a_obj_flag.h
```diff
@@ -18,19 +28,141 @@ class FlagJoint_c {
...
+    /* 0x570 */ request_of_phase_process_class mPhase;
+    /* 0x578 */ request_of_phase_process_class mPhase2;
+    /* 0x580 */ FlagJoint_c mFlagJoints[4];
+    /* 0x5d0 */ cXyz mPos;
+    /* 0x5dc */ float field_0x5dc;
+    /* 0x5e0 */ s16 field_0x5e0;
+    /* 0x5e2 */ short field_0x5e2;
+    /* 0x5e4 */ short field_0x5e4;
+    /* 0x5e6 */ short field_0x5e6;
+    /* 0x5e8 */ short field_0x5e8;
```
TakaRikka, line 43: f32 and s16 here too

include/d/actor/d_a_obj_flag.h
```diff
@@ -18,19 +28,141 @@ class FlagJoint_c {
...
+
+        dComIfGd_setList();
+
+        return 1;
+    }
+
+    int createHeap() {
+        bool tmp = 0;
+        char angle = (u8)shape_angle.x;
+        if(angle <= -1 || angle > 'c') {
```
TakaRikka, line 108: this is probably supposed to be a u8, char doesn't make much sense in the context of what its doing

include/d/actor/d_a_obj_flag.h
```diff
@@ -18,19 +28,141 @@ class FlagJoint_c {
...
+        /* 0x1c */ short field_0x1c;
+        /* 0x1e */ short field_0x1e;
+        /* 0x20 */ short field_0x20;
+        /* 0x22 */ short field_0x22;
+        /* 0x24 */ u8 field_0x24;
+        /* 0x25 */ u8 field_0x25;
+        /* 0x26 */ u8 field_0x26;
+        /* 0x28 */ float field_0x28;
+        /* 0x2c */ float field_0x2c;
+        /* 0x30 */ float field_0x30;
```
TakaRikka, line 161: f32 and s16 here

src/d/actor/d_a_obj_flag.cpp
```diff
@@ -6,235 +6,214 @@
...
+    mPos = cXyz(current.pos.x, current.pos.y + field_0x5dc, current.pos.z);
+    mFlagJoints[0].rv = (short)(int)(cM_rnd() * 65535.0f);
+    mFlagJoints[1].rv = (short)(int)(cM_rnd() * 65535.0f);
+    mFlagJoints[2].rv = mFlagJoints[1].rv;
+    mFlagJoints[3].rv = mFlagJoints[2].rv;
+
+    field_0x5e2 = (short)(int)(cM_rnd() * 65535.0f);
+    field_0x5e4 = (short)(int)(cM_rnd() * 65535.0f);
+    field_0x5e6 = (short)(int)(cM_rnd() * 65535.0f);
+    field_0x5e8 = (short)(int)(cM_rnd() * 65535.0f);
```
TakaRikka, line 39: are these double casts necessary? does implicit type conversion work instead?

src/d/actor/d_a_obj_flag.cpp
```diff
@@ -6,235 +6,214 @@
...
 void daObjFlag_c::initBaseMtx() {
-    // NONMATCHING
+    mDoMtx_stack_c::transS(current.pos);
+    PSMTXCopy(mDoMtx_stack_c::now, mpModel2->mBaseTransformMtx);
+    cullMtx = mpModel2->mBaseTransformMtx;
+    if (mpModel1 != NULL) {
+        mDoMtx_stack_c::transS(mPos);
+        mDoMtx_YrotM(mDoMtx_stack_c::now, shape_angle.y);
+        PSMTXCopy(mDoMtx_stack_c::now, mpModel1->mBaseTransformMtx);
+        cullMtx = mpModel1->mBaseTransformMtx;
```
TakaRikka, line 53: this can probably be improved with inlines like setBaseTRMtx, fopAcM_SetMtx, and mDoMtx_stack_c::YrotM

src/d/actor/d_a_obj_flag.cpp
```diff
@@ -6,235 +6,214 @@
...
-
 /* 80BEB9AC-80BEBC58 0002AC 02AC+00 1/1 0/0 0/0 .text            calcJointAngle__11daObjFlag_cFv */
 void daObjFlag_c::calcJointAngle() {
-    // NONMATCHING
+    cXyz* direction;
+    float power;
+
+    dKyw_get_AllWind_vec(&mPos, direction, &power);
+    if(power > 0.0f) {
+        Z2GetAudioMgr()->seStartLevel(JAISoundID(Z2SE_OBJ_FLAG_TRAILING), &mPos, power * 127.0f, 0, 1.0, 1.0, -1.0, -1.0, 0);
```
TakaRikka, line 70: the explicit JAISoundID creation isnt necessary, you can just pass in the sound enum and it'll automatically convert it thanks to the u32 operator

src/d/actor/d_a_obj_flag.cpp
```diff
@@ -6,235 +6,214 @@
...
+    if(swing > 4000.0f) {
+        swing = 4000.0f;
+    }
+    if(param_0 == 0) {
+        return swing;
+    }
+
+    float var3 = (float)(int)attr().field_0x14 * cM_ssin(field_0x5e2);
+    float var4 = (float)(int)attr().field_0x16 * cM_ssin(field_0x5e4);
+    float var5 = (float)(int)attr().field_0x18 * cM_ssin(field_0x5e6);
```
TakaRikka, line 153: same thing about if removing some of these casts can work

include/d/actor/d_a_obj_flag.h
tunamayo04: yes I tried with char but with a char it gets rid of a clrlwi that I could only get to show up by casting to u8

TakaRikka: the 'c' should at least be changed into 99 instead, since the value isnt supposed to be treated as a character but rather a numerical value

TakaRikka (approved):

PR #2353, opened 2025-03-27
d_a_cow Equivalent

soruh:
Let me know if I should change anything or if I missed something to get this `Matching`!

### DIFF
--- a/include/d/d_com_inf_actor.h
+++ b/include/d/d_com_inf_actor.h
@@ -5,7 +5,7 @@ class fopAc_ac_c;
 
 class dComIfGoat_info_c {
 public:
-    /* 0x0 */ fopAc_ac_c* mGoatActor;
+    /* 0x0 */ fopAc_ac_c* mThrowActor;
 };
 
 class dComIfAc_info_c {
@@ -22,11 +22,11 @@ inline bool dComIfA_PauseCheck() {
 }
 
 inline fopAc_ac_c* dComIfGoat_GetThrow() {
-    return g_dComIfGoat_gameInfo.mGoatActor;
+    return g_dComIfGoat_gameInfo.mThrowActor;
 }
 
 inline void dComIfGoat_SetThrow(fopAc_ac_c* actor) {
-    g_dComIfGoat_gameInfo.mGoatActor = actor;
+    g_dComIfGoat_gameInfo.mThrowActor = actor;
 }
 
 #endif /* D_COM_D_COM_INF_ACTOR_H */
--- a/include/d/d_com_inf_actor.h
+++ b/include/d/d_com_inf_actor.h
@@ -5,7 +5,7 @@ class fopAc_ac_c;
 
 class dComIfGoat_info_c {
 public:
-    /* 0x0 */ fopAc_ac_c* field_0x0;
+    /* 0x0 */ fopAc_ac_c* mGoatActor;
 };
 
 class dComIfAc_info_c {
@@ -22,7 +22,11 @@ inline bool dComIfA_PauseCheck() {
 }
 
 inline fopAc_ac_c* dComIfGoat_GetThrow() {
-    return g_dComIfGoat_gameInfo.field_0x0;
+    return g_dComIfGoat_gameInfo.mGoatActor;
+}
+
+inline void dComIfGoat_SetThrow(fopAc_ac_c* actor) {
+    g_dComIfGoat_gameInfo.mGoatActor = actor;
 }
 
 #endif /* D_COM_D_COM_INF_ACTOR_H */
--- a/src/d/actor/d_a_cow.cpp
+++ b/src/d/actor/d_a_cow.cpp
@@ -70,35 +70,35 @@ extern fpc_ProcID gWolfBustersID[N_WOLF_BUSTERS] = {-1, -1, -1};
 /* 806585CC-80658730 0000EC 0164+00 6/6 0/0 0/0 .text            calcRunAnime__7daCow_cFi */
 int daCow_c::calcRunAnime(int resetRunType) {
     if (resetRunType) {
-        mRunType = daCow_RunType_None_e;
+        mRunType = daCow_c::RunType_None;
     }
 
     switch (mRunType) {
-    case daCow_RunType_None_e:
+    case daCow_c::RunType_None:
         if (speedF < 35.0f) {
-            setBck(daCow_Animation_Walk_e, J3DFrameCtrl::EMode_LOOP, 5.0f, 1.0f);
-            mRunType = daCow_RunType_Walk_e;
+            setBck(daCow_c::Animation_Walk, J3DFrameCtrl::EMode_LOOP, 5.0f, 1.0f);
+            mRunType = daCow_c::RunType_Walk;
         } else {
-            setBck(daCow_Animation_Run_e, J3DFrameCtrl::EMode_LOOP, 5.0f, 1.0f);
-            mRunType = daCow_RunType_Run_e;
+            setBck(daCow_c::Animation_Run, J3DFrameCtrl::EMode_LOOP, 5.0f, 1.0f);
+            mRunType = daCow_c::RunType_Run;
         }
         break;
-    case daCow_RunType_Walk_e: {
+    case daCow_c::RunType_Walk: {
         f32 animationSpeed = (speedF * 3.0f) / 25.0f;
         if (animationSpeed > 5.0f) {
             animationSpeed = 5.0f;
         }
         mpMorf->setPlaySpeed(animationSpeed);
         if (speedF > 35.0f) {
-            setBck(daCow_Animation_Run_e, J3DFrameCtrl::EMode_LOOP, 5.0f, 1.0f);
-            mRunType = daCow_RunType_Run_e;
+            setBck(daCow_c::Animation_Run, J3DFrameCtrl::EMode_LOOP, 5.0f, 1.0f);
+            mRunType = daCow_c::RunType_Run;
         }
     } break;
-    case daCow_RunType_Run_e:
+    case daCow_c::RunType_Run:
         mpMorf->setPlaySpeed(1.3f);
         if (speedF < 35.0f) {
-            setBck(daCow_Animation_Walk_e, J3DFrameCtrl::EMode_LOOP, 5.0f, 1.0f);
-            mRunType = daCow_RunType_Walk_e;
+            setBck(daCow_c::Animation_Walk, J3DFrameCtrl::EMode_LOOP, 5.0f, 1.0f);
+            mRunType = daCow_c::RunType_Walk;
         }
         break;
     }
@@ -218,8 +218,8 @@ void daCow_c::setActetcStatus() {
 
 /* 80658C78-80658CA4 000798 002C+00 3/3 0/0 0/0 .text            checkNadeNadeFinish__7daCow_cFv */
 bool daCow_c::checkNadeNadeFinish() {
-    if (GET_FLAG(daCow_Flag_NaderuFinish_e)) {
-        CLEAR_FLAG(daCow_Flag_NaderuFinish_e);
+    if (GET_FLAG(daCow_c::Flag_NaderuFinish)) {
+        CLEAR_FLAG(daCow_c::Flag_NaderuFinish);
         mNadeNade = false;
         return true;
     }
@@ -228,8 +228,8 @@ bool daCow_c::checkNadeNadeFinish() {
 
 /* 80658CA4-80658CD0 0007C4 002C+00 5/5 0/0 0/0 .text            checkNadeNade__7daCow_cFv */
 bool daCow_c::checkNadeNade() {
-    if (GET_FLAG(daCow_Flag_Naderu_e)) {
-        CLEAR_FLAG(daCow_Flag_Naderu_e);
+    if (GET_FLAG(daCow_c::Flag_Naderu)) {
+        CLEAR_FLAG(daCow_c::Flag_Naderu);
         mNadeNade = true;
         return 1;
     }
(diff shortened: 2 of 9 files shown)


src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+                speedF = 0.0f;
+                setProcess(&daCow_c::action_wait, false);
+            }
+        }
+    }
+
+    mSph[0].ClrTgHit();
+    mSph[1].ClrTgHit();
+    mSph[2].ClrTgHit();
+    STATIC_ASSERT(N_COW_COLLIDERS == 3);
```
TakaRikka, line 396: this doesnt seem like it exists in the debug rom? so it should probably be removed

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
 }
 
 /* 8065945C-80659540 000F7C 00E4+00 1/1 0/0 0/0 .text            setEnterCow20__7daCow_cFv */
 void daCow_c::setEnterCow20() {
-    // NONMATCHING
+    for (int iCow = 0; iCow < 20; iCow++) {
+        cXyz spawnPosition(l_CowRoomPosX[iCow], l_CowRoomPosY, l_CowRoomPosZ[iCow & 1]);
+
+        cLib_onBit<u32>(l_CowRoomNo, 1 << iCow);
```
TakaRikka, line 404: the clib_onbit call isnt in the debug rom, so this should be done raw

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+
+        csXyz spawnAngle;
+        if (iCow & 1) {
+            spawnAngle.set(0, 0, 0);
+        } else {
+            spawnAngle.set(0, -0x8000, 0);
+        }
+
+        fopAcM_create(0x106, ~0xfb, &spawnPosition, fopAcM_GetRoomNo(this), &spawnAngle, 0, -1);
```
TakaRikka, line 413: since this is passing in an actor parameter, it's preferred to write it like `0xFFFFFF04` rather than `~0xFB`

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+
+#define GET_FLAG(mask) (mFlags & (mask))    // cLib_checkBit<u16>(mFlags, (mask))
+#define CLEAR_FLAG(mask) mFlags &= ~(mask)  // cLib_offBit<u16>(mFlags, (mask))
+#define TICK_TIMER(name)                                                                           \
+    if ((name)) {                                                                                  \
+        (name)--;                                                                                  \
+    }
+#define COW_INDEX (u8)(fopAcM_GetParam(this) >> 8)
+#define CLAMP(val, min, max)                                                                       \
+    STATIC_ASSERT(min <= max);                                                                     \
```
TakaRikka, line 30: this doesn't seem to in the debug rom? so it can be removed

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+        xMax = 100.0f;
+    }
+    mDoMtx_stack_c::YrotS(-shape_angle.y);
+    mDoMtx_stack_c::transM(-current.pos.x, -current.pos.y, -current.pos.z);
+
+    Vec carryPosition;
+    mDoMtx_stack_c::multVec(&daPy_getPlayerActorClass()->current.pos, &carryPosition);
+    if (fabsf(carryPosition.x) < xMax && carryPosition.z > 0.0f && carryPosition.z < zMax) {
+        // todo: what does this mean
+        cLib_onBit<u32>(attention_info.flags, 0x10);  // in debug this is 0x80
```
TakaRikka, line 203: this isnt in the debug rom, so it should be done raw

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+    }
 }
 
 /* 80658C18-80658C78 000738 0060+00 5/5 0/0 0/0 .text            setActetcStatus__7daCow_cFv */
 void daCow_c::setActetcStatus() {
-    // NONMATCHING
+    if (!mNadeNade) {
+        s32 playerAngle = fopAcM_seenPlayerAngleY(this);
+        if (playerAngle < 0x6000 && playerAngle > 0x2000) {
+            cLib_onBit<u32>(attention_info.flags, 0x80);  // in debug this is 0x800
```
TakaRikka, line 212: this isnt in the debug rom, so it should be done raw

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
-    (void*)func_80662D60,
-};
-
-/* 80663440-8066344C 0004E8 000C+00 2/2 0/0 0/0 .data            __vt__12J3DFrameCtrl */
-SECTION_DATA extern void* __vt__12J3DFrameCtrl[3] = {
-    (void*)NULL /* RTTI */,
-    (void*)NULL,
-    (void*)__dt__12J3DFrameCtrlFv,
-};
+        cLib_onBit<u32>(l_CowRoomNo, 1 << cowNumber);
```
TakaRikka, line 428: this isnt in the debug rom, so it should be done raw

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
-    (void*)__dt__8cM3dGPlaFv,
-};
+        csXyz spawnAngle;
+        if (cowNumber & 1) {
+            spawnAngle.set(0, 0, 0);
+        } else {
+            spawnAngle.set(0, -0x8000, 0);
+        }
+        int roomNumber = fopAcM_GetRoomNo(this);
+        fopAcM_create(0x106, ~0xfb, &spawnPosition, roomNumber, &spawnAngle, 0, -1);
```
TakaRikka, line 437: same thing about actor parameters

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+    mIntersectedPlanes = 0;
+    if (planeTri[0]) {
+        cLib_onBit<u8>(mIntersectedPlanes, 1);
+    }
+    if (planeTri[1]) {
+        cLib_onBit<u8>(mIntersectedPlanes, 2);
+    }
+    if (planeTri[2]) {
+        cLib_onBit<u8>(mIntersectedPlanes, 4);
+    }
```
TakaRikka, line 1009: none of these clib_onbit calls are in the debug rom, so should be done raw

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+        mSph[iSphere].OffCoSetBit();
+        mSph[iSphere].OnCoNoCrrBit();
+        mCcStts.ClrCcMove();
+    }
+    mAcchCir.SetWall(0.0f, 0.0f);
+
+    u8 cowIndex = cM_rndF(20.0f);
+    if (cLib_checkBit<u32>(l_CowRoomNo, 1 << cowIndex)) {
+        for (int iCow = 0; iCow < 20; iCow++) {
+            if (!cLib_checkBit<u32>(l_CowRoomNo, 1 << iCow)) {
```
TakaRikka, line 1419: none of these clib_checkbit calls are in the debug rom, so should be done raw

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+    }
+
+    if (cowIndex >= 20) {
+        cowIndex = 19;
+    }
+
+    current.pos.x = l_CowRoomPosX[cowIndex];
+    current.pos.z = l_CowRoomPosZ[cowIndex & 1];
+    old.pos = current.pos;
+    cLib_onBit<u32>(l_CowRoomNo, 1 << cowIndex);
```
TakaRikka, line 1433: doesnt exist in debug, should be done raw

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+        mSound.startCreatureVoice(Z2SE_GOAT_V_ANGRY, -1);
+        mDealDamageTimer = 0;
+        break;
+    case daCow_c::Mode_1:
+        setSeSnort();
+
+        if (!player->checkHorseRide()) {
+            cLib_onBit<u32>(attention_info.flags, 1);
+        } else {
+            cLib_offBit<u32>(attention_info.flags, 1);
```
TakaRikka, line 1722: doesnt exist in debug, should be done raw

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+                } else {
+                    if (playerDistance < 500.0f) {
+                        mForgetCowPTimer = 35;
+                    } else {
+                        if ((playerDistance >= 1500.0f) && angleToPlayer >= 0x5800) {
+                            current.angle.y = mSavedAngle.y;
+                            mCrazy = daCow_c::Crazy_BeforeCatch;
+                            return;
+                        }
+                    }
```
TakaRikka, line 1853: using here as an example but its present in many places in this TU, many of these if else chains can be simplified into if else if's that reduce the amount of nesting

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+
+    cXyz aruPos = aru->current.pos;
+    s16 aruAngle = cLib_targetAngleY(&current.pos, &aru->current.pos);
+
+    switch (mMode) {
+    case daCow_c::Mode_0:
+        mMode = daCow_c::Mode_1;
+        mCrazy = daCow_c::Crazy_Wait;
+        calcRunAnime(true);
+        cLib_onBit<u32>(attention_info.flags, 1);
```
TakaRikka, line 2813: doesnt exist in debug, should be done raw

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
-#pragma pop
-
 /* 806615EC-80661720 00910C 0134+00 1/1 0/0 0/0 .text            setAttnPos__7daCow_cFv */
 void daCow_c::setAttnPos() {
-    // NONMATCHING
+    cXyz arg;
+    cXyz pos;
+
+    if (mpMorf) {
+        if (cLib_checkBit<u32>(attention_info.flags, 0x80)) {
```
TakaRikka, line 2995: doesnt exist in debug, should be done raw

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
-void daCow_c::createHeapCallBack(fopAc_ac_c* param_0) {
-    // NONMATCHING
-}
+#ifdef DEBUG
+    if (modelData == NULL) {
+        JUTAssertion::showAssert(JUTAssertion::getSDevice(), "d_a_cow.cpp", 0xef2,
+                                 "0 != modelData");
+        OSPanic("d_a_cow.cpp", 0xef2, "Halt");
+    }
+#endif
```
TakaRikka, line 3104: this should just be a JUT_ASSERT

src/d/actor/d_a_cow.cpp
```diff
@@ -1,2589 +1,3449 @@
...
+
+    f32 rand = cM_rnd();
+    u8 iSpeed = 0;
+    if (rand < 0.1f) {
+        if (!cLib_checkBit<u32>(l_CowType, 1)) {
+            l_CowType |= 1;
+            iSpeed = 1;
+        }
+    } else {
+        if (rand >= 0.9f && !cLib_checkBit<u32>(l_CowType, 2)) {
```
TakaRikka, line 3224: doesn't exist in debug rom, should be done raw

soruh:
I have removed all of the extra inlines not in the debug rom that I missed before. I goot too much into the "write it in a reasonable way" groove. I'll pay more attention to use exactly the inlines that are in the debug rom in the future.

src/d/actor/d_a_cow.cpp
soruh: Done, hope I used it correctly

src/d/actor/d_a_cow.cpp
soruh: I have changed this in the places I found (except for where it made more sense)

src/d/actor/d_a_cow.cpp
soruh: I have changed this. Is there some insight to how there parameters work/why the non inverted way makes more sense?

include/d/d_com_inf_actor.h
```diff
@@ -5,7 +5,7 @@ class fopAc_ac_c;
 
 class dComIfGoat_info_c {
 public:
-    /* 0x0 */ fopAc_ac_c* field_0x0;
+    /* 0x0 */ fopAc_ac_c* mGoatActor;
```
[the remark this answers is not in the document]
soruh: done

src/d/actor/d_a_cow.cpp
soruh: Should I remove the whole macro or just the assert. Neither would appear in the debug rom but this code is repeated quite a few times...

src/d/actor/d_a_cow.cpp
TakaRikka: its basically just a 32bit space to store various parameters for the actor to use, often times splitting up each byte of the word as a separate parameter but sometimes splitting it even further. thus it's easier to read the parameter fields when they're written this way (0xFFFFFF04 & 0xFF is 4, whereas its not as intuitive reading it the inverted way)

src/d/actor/d_a_cow.cpp
TakaRikka: the macro itself is probably fine, i dont have a strong opinion either way on it, but the static assert should be removed

src/d/actor/d_a_cow.cpp
soruh: Makes sense, thank you

src/d/actor/d_a_cow.cpp
soruh: great, I did that in the previous commit already

src/d/actor/d_a_cow.cpp
```diff
@@ -3095,13 +3065,7 @@ static int daCow_Execute(void* actor) {
...
     J3DModelData* modelData = static_cast<J3DModelData*>(dComIfG_getObjectRes("Cow", 0x1f));
 
-#ifdef DEBUG
-    if (modelData == NULL) {
-        JUTAssertion::showAssert(JUTAssertion::getSDevice(), "d_a_cow.cpp", 0xef2,
-                                 "0 != modelData");
-        OSPanic("d_a_cow.cpp", 0xef2, "Halt");
-    }
-#endif
+    JUT_ASSERT(modelData, 3826);
```
TakaRikka, line 3068: just one minor thing, these params are backwards. line number comes before the expression

soruh: Yhey sure were. I even looked at the header and then messed it up anyways...
Fixed now

TakaRikka (approved):

soruh:
thanks for the review!

PR #2362, opened 2025-03-31
d_a_obj_dust work, 90% matching

vladzhdanov:
Pushing my work on `d_a_obj_dust `. Its currently 90% matching, with the only differences from equivalence being with the `sinShort` inlining in ` daObjDust_c::Execute` and some [extra 0 padding](https://i.imgur.com/v5ftUpK.png) in `.rodata` which I haven't been able to figure out.

### DIFF
--- a/src/d/actor/d_a_obj_dust.cpp
+++ b/src/d/actor/d_a_obj_dust.cpp
@@ -4,97 +4,9 @@
 */
 
 #include "d/actor/d_a_obj_dust.h"
-#include "dol2asm.h"
 
-
-//
-// Forward References:
-//
-
-extern "C" void Search_Ymb__11daObjDust_cFv();
-extern "C" void RideOn_Angle__11daObjDust_cFRsfsf();
-extern "C" void Check_RideOn__11daObjDust_cFv();
-extern "C" void initBaseMtx__11daObjDust_cFv();
-extern "C" void setBaseMtx__11daObjDust_cFv();
-extern "C" static void rideCallBack__FP4dBgWP10fopAc_ac_cP10fopAc_ac_c();
-extern "C" static void daObjDust_Draw__FP11daObjDust_c();
-extern "C" static void daObjDust_Execute__FP11daObjDust_c();
-extern "C" static bool daObjDust_IsDelete__FP11daObjDust_c();
-extern "C" static void daObjDust_Delete__FP11daObjDust_c();
-extern "C" static void daObjDust_Create__FP10fopAc_ac_c();
-extern "C" void __dt__12dBgS_ObjAcchFv();
-extern "C" void CreateHeap__11daObjDust_cFv();
-extern "C" void Create__11daObjDust_cFv();
-extern "C" void Execute__11daObjDust_cFPPA3_A4_f();
-extern "C" void Draw__11daObjDust_cFv();
-extern "C" void Delete__11daObjDust_cFv();
-extern "C" void func_80BE2F74(void* _this, s16);
-extern "C" static void func_80BE2F80();
-extern "C" static void func_80BE2F88();
-extern "C" extern char const* const d_a_obj_dust__stringBase0;
-
-//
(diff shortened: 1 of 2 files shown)


src/d/actor/d_a_obj_dust.cpp
```diff
@@ -105,407 +17,269 @@ SECTION_RODATA static u8 const lit_3673[4 + 4 /* padding */] = {
...
-static u8 data_80BE30F8[4];
+/* 80BE3054-80BE3058 -00001 0004+00 3/3 0/0 0/0 .data            l_arcName */
+static const char* l_arcName = "M_Dust";
+
+/* 80BE30F8-80BE30FC 000000 0004+00 2/2 0/0 0/0 .bss             e_ymb__26@unnamed@d_a_obj_dust_cpp@ */
+static fopAc_ac_c* s_ymb;
 
 /* 80BE30FC-80BE3100 000004 0004+00 2/2 0/0 0/0 .bss e_ymb_Pos__26@unnamed@d_a_obj_dust_cpp@ */
-static u8 data_80BE30FC[4];
+static cXyz* s_ymb_Pos;
```
TakaRikka, line 28: these names look wrong, should be e_ymb, not s_ymb. also the `@unnamed@d_a_obj_dust_cpp@` means they should be wrapped in an anonymous namespace

src/d/actor/d_a_obj_dust.cpp
```diff
@@ -105,407 +17,269 @@ SECTION_RODATA static u8 const lit_3673[4 + 4 /* padding */] = {
...
+    if (std::abs(i_rideActor->current.pos.y - i_rideActor->old.pos.y) > 30.0f) {
+        static_cast<daObjDust_c*>(i_bgActor)->field_0x5a8 = i_rideActor->current.pos.y - i_rideActor->old.pos.y;
+    }
 }
 
 /* 80BE27BC-80BE27E8 00057C 002C+00 1/0 0/0 0/0 .text            daObjDust_Draw__FP11daObjDust_c */
-static void daObjDust_Draw(daObjDust_c* param_0) {
-    // NONMATCHING
+static void daObjDust_Draw(daObjDust_c* i_this) {
+    i_this->MoveBGDraw();
```
TakaRikka, line 115: all actor methods (create, delete, isDelete, execute, draw) should return int values, so they should return in the inner function call

src/d/actor/d_a_obj_dust.cpp
```diff
@@ -105,407 +17,269 @@ SECTION_RODATA static u8 const lit_3673[4 + 4 /* padding */] = {
...
-void daObjDust_c::Execute(f32 (**param_0)[3][4]) {
-    // NONMATCHING
+// NONMATCHING somehow this needs to not inline TSinCosTable<13,f32>::sinShort
+int daObjDust_c::Execute(Mtx** i_mtx) {
+    fopAcM_seStartLevel(this, Z2SE_OBJ_FLOATBOARD, 0);
+    daPy_py_c* pdVar = daPy_getPlayerActorClass();
+    fopAcM_GetPosition(pdVar);
+    fopAcM_GetSpeed(pdVar);
+
+    if (fopAcM_SearchByName(PROC_E_YMB, (fopAc_ac_c**) &s_ymb) != 0 && s_ymb != NULL) {
```
TakaRikka, line 204: i dont think the cast should be needed here?

src/d/actor/d_a_obj_dust.cpp
vladzhdanov: Ah yeah, I changed it from `e_ymb` since it was static. Updated now.

src/d/actor/d_a_obj_dust.cpp
vladzhdanov: Yep, you're right. Updated

TakaRikka (approved):

PR #2364, opened 2025-03-31
d_a_ep work

carter-ktb21:


src/d/actor/d_a_ep.cpp
```diff
@@ -1,1164 +1,965 @@
...
+        mDoAud_seStart(Z2SE_CM_BODYFALL_WATER_S, &i_this->field_0x4, 0, 0);
+    }
+
+    return rv;
 }
 
 /* 80468A70-80468A90 0008F0 0020+00 1/1 0/0 0/0 .text hahen_normal__FP8ep_classP10ep_hahen_s */
-static void hahen_normal(ep_class* param_0, ep_hahen_s* param_1) {
-    // NONMATCHING
+static void hahen_normal(ep_class* e_this, ep_hahen_s* i_this) {
```
TakaRikka, line 147: the use of `e_this` and `i_this` throughout the whole TU should be cleaned up to use `i_this` for `ep_class` and `hahen_s` for `ep_hahen_s`.

src/d/actor/d_a_ep.cpp
```diff
@@ -1,1164 +1,965 @@
...
+        0x0100, 0x8110,
+    };
+    static u16 l_particle_fire_B[2] = {
+        0x0101, 0x8111
+    };
+    static u16 l_particle_kagerou[2] = {
+        0x0103, 0x8112
+    };
+
+    cXyz local_1c = cXyz(e_this->field_0x634.x, e_this->field_0x634.y + -240.0f + 235.0f +15.0f, e_this->field_0x634.z);
```
TakaRikka, line 432: should just be `cXyz sp1C(...`

src/d/actor/d_a_ep.cpp
```diff
@@ -1,1164 +1,965 @@
...
+    e_this->mLightInf.mFluctuation = 1.0f;
+    e_this->scale.y = e_this->field_0x5b0;
+
+    if (e_this->scale.y > 0.5f) {
+        e_this->mSph1.OnAtSetBit();
+    } else {
+        e_this->mSph1.OffAtSetBit();
+    }
+
+    e_this->mSph1.SetC(cXyz(e_this->field_0x634.x, e_this->field_0x634.y + 30.0f, e_this->field_0x634.z));
```
TakaRikka, line 553: this cXyz should be put on the stack since it doesn't call a destructor immediately after

src/d/actor/d_a_ep.cpp
```diff
@@ -1,1164 +1,965 @@
...
+            cMtx_YrotS(*calc_mtx, e_this->field_0xa58);
+
+            cXyz local_34(0.0f, 1.0f, e_this->field_0xa54);
+            cXyz cStack_40;
+            MtxPosition(&local_34, &cStack_40);
+
+            JGeometry::TVec3<float> aTStack_88;
+            aTStack_88 = cStack_40;
+            JPABaseEmitter* emitter = (JPABaseEmitter*)e_this->field_0x610;
+            emitter->setDirection(aTStack_88);
```
TakaRikka, line 575: you shouldnt need to explicitly create a TVec3 here, just passing in a cXyz should implicitly do the conversion. also emitter shouldnt be stored to a temp var, you can just call the member var directly

src/d/actor/d_a_ep.cpp
```diff
@@ -1,1164 +1,965 @@
...
+
+            if (e_this->field_0xa50 == 1) {
+                emitter = (JPABaseEmitter*)e_this->field_0x610;
+                emitter->becomeInvalidEmitter();
```
TakaRikka, line 579: same thing here, no temp needed

src/d/actor/d_a_ep.cpp
```diff
@@ -1,1164 +1,965 @@
...
+    if (e_this->field_0xa79) {
+        hahen_move(e_this);
+    }
+
+    return 1;
 }
 
 /* 8046A6D4-8046A6DC 002554 0008+00 1/0 0/0 0/0 .text            daEp_IsDelete__FP8ep_class */
-static bool daEp_IsDelete(ep_class* param_0) {
+static bool daEp_IsDelete(ep_class* e_this) {
```
TakaRikka, line 768: returns int

src/d/actor/d_a_ep.cpp
```diff
@@ -1,1164 +1,965 @@
...
+        JUT_ASSERT(0x884, modelData != 0);
+
+        for (int i = 0; i < 6; i++) {
+            e_this->mHahen[i].mpModel = mDoExt_J3DModel__create(modelData, 0x80000, 0x11000084);
+            if (e_this->mHahen[i].mpModel == NULL) {
+                return 0;
+            }
+        }
+        rv = 1;
+    }
```
TakaRikka, line 810: using here as an example but probably in many other places in this TU, the rv temp isn't necessary. just return the value directly, and it'll also let you remove the unnecessary else block nesting

src/d/actor/d_a_ep.cpp
```diff
@@ -1,1164 +1,965 @@
...
 
 /* 8046A858-8046A8D4 0026D8 007C+00 1/1 0/0 0/0 .text            daEp_CreateInit__FP10fopAc_ac_c */
-static void daEp_CreateInit(fopAc_ac_c* param_0) {
-    // NONMATCHING
+static void daEp_CreateInit(fopAc_ac_c* a_this) {
+    ep_class* e_this = (ep_class*)a_this;
+
+    e_this->field_0x5a4 = 0;
+    e_this->field_0x60d = 0;
+    e_this->cullMtx = e_this->field_0x570;
```
TakaRikka, line 821: fopAcM_SetMtx

src/d/actor/d_a_ep.cpp
```diff
@@ -1,1164 +1,965 @@
...
+        i_this->mSph1.Set(sph_src);
+        i_this->mSph1.SetStts(&i_this->mStts);
+
+        if (i_this->field_0x60c == 1) {
+            i_this->mSph1.SetR(70.0f);
+        }
+
+        if (i_this->field_0xa5a < 0) {
+            #ifdef DEBUG
+            OSReport("   /////   EP PARAM %x\n");
```
TakaRikka, line 900: you can use the OS_REPORT macro to avoid having to wrap OSReport calls in ifdefs

src/d/actor/d_a_ep.cpp
carter-ktb21: I went through and got rid of rv where it wasn't needed to match

PR #2374, opened 2025-04-05
d_a_npc_shop0 matching

vladzhdanov:
Got `d_a_npc_shop0 ` fully matching!

### DIFF
--- a/src/d/actor/d_a_npc_shop0.cpp
+++ b/src/d/actor/d_a_npc_shop0.cpp
@@ -4,190 +4,385 @@
 */
 
 #include "d/actor/d_a_npc_shop0.h"
-#include "dol2asm.h"
-
-//
-// Forward References:
-//
-
-extern "C" static void createHeapCallBack__FP10fopAc_ac_c();
-extern "C" void __dt__13daNpc_Shop0_cFv();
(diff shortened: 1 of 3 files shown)


src/d/actor/d_a_npc_shop0.cpp
```diff
@@ -4,617 +4,402 @@
...
-/* 80AEBC54-80AEBC54 000080 0000+00 0/0 0/0 0/0 .rodata          @stringBase0 */
-#pragma push
-#pragma force_active on
-SECTION_DEAD static char const* const stringBase_80AEBC6F = ".bck";
-#pragma pop
-
 /* 80AEBDC0-80AEBDD8 000000 0015+03 1/1 0/0 0/0 .bss             l_fileName */
-static u8 l_fileName[21 + 3 /* padding */];
+static char l_fileName[21];
+static u8 gap_05_00000015_bss[3]; // .bss padding
```
TakaRikka, line 199: is this actually necessary for matching? i'd assume that this is just compiler generated

vladzhdanov: Oh yep you're right, it still matches with it removed. Updated

hatal175 (approved):

PR #2378, opened 2025-04-07
Debug JStudio fix + match one retail func

LagoLunatic:
The weird gameinfo load in daAlink_c::getDemoLookActor was caused by the redundant cast, some callers need the extra cast and some don't

### DIFF
--- a/config/ShieldD/splits.txt
+++ b/config/ShieldD/splits.txt
@@ -1996,57 +1996,57 @@ JSystem/JStage/JSGObject.cpp:
 	.data       start:0x806F13B8 end:0x806F13F8
 	.sdata2     start:0x80751868 end:0x80751870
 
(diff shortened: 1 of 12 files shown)


TakaRikka:
the fake header stuff is unfortunate cause its pretty ugly, but if its the only way to support both versions then i guess we'll have to live with it

LagoLunatic:
> the fake header stuff is unfortunate cause its pretty ugly, but if its the only way to support both versions then i guess we'll have to live with it

Maybe someone who knows more about C++ templates could find a better way, I still don't really understand *why* the newer compiler version refuses to let the template class be used after it's defined which seems pretty backwards. But yeah I couldn't think of an alternative

TakaRikka (approved):

PR #2390, opened 2025-04-12
Debug work for d_a_e_rdy to get 0.10% GCN progress

YunataSavior:
Using the debug build, I attempted to solve d_a_e_rdy's regalloc issues. I managed to get some functions to match, and demo_camera matches more closely, but some regalloc issues remain. For the functions which continue to have regalloc issues, they mostly match on debug, so IDK what else can be done.

These C-style TUs are very frustrating....


TakaRikka (commented):
overall looks fine, only thing is for debug differences that aren't related to debug features, i'd use a `#if VERSION != VERSION_SHIELD_DEBUG` check instead since we'll want to have a debug define that only changes whether debug features are enabled or not, not version differences

hatal175 (approved):

PR #2398, opened 2025-04-16
Daswc

hatal175:


configure.py
```diff
@@ -312,10 +312,12 @@
 # REL flags
 cflags_rel = [
     *cflags_framework,
-    "-sdata 0",
-    "-sdata2 0",
 ]
 
+# I think that debug version doesn't pass these flags but the build fails on a relocation
```
TakaRikka, line 317: whats the reasoning for this thought? it seems unlikely based on what we know about rels for this to be true, but maybe theres a similar thing that needs to be adjusted for correct behavior

hatal175:
Just take a look at any rel. I think you'll see most don't have an rodata
section and in most you'll see mismatches in sdata and sdata2 locations.

hatal175:
Of course, my sampling might be wrong, so a second look is welcome.

TakaRikka:
i think the problem here is that in the shield version, some of the rels were instead just made into dol TUs, like d_a_grass and d_a_swc, etc. meaning d_a_swc would be compiled with dol flags for shield, but rel flags for GC

hatal175:
Hmm. I'll adjust the pool request.

TakaRikka (approved):

PR #2400, opened 2025-04-18
d_a_npc_uri almost matching

hev7:
https://github.com/zeldaret/tp/issues/725

### DIFF
--- a/include/d/actor/d_a_npc_uri.h
+++ b/include/d/actor/d_a_npc_uri.h
@@ -11,80 +11,179 @@
  * @details
  *
  */
-class daNpc_Uri_c : public fopAc_ac_c {
+class daNpc_Uri_c : public daNpcT_c {
 public:
+    typedef int (daNpc_Uri_c::*cutFunc)(int);
+    typedef int (daNpc_Uri_c::*actionFunc)(void*);
+
+    enum Type {
+        TYPE_0,
+        TYPE_1,
+        TYPE_2,
+        TYPE_3,
+        TYPE_4,
+        TYPE_5,
+        TYPE_6,
+    };
+
     /* 80B267AC */ ~daNpc_Uri_c();
-    /* 80B268F8 */ void create();
-    /* 80B26BE4 */ void CreateHeap();
-    /* 80B2708C */ void Delete();
-    /* 80B270C0 */ void Execute();
-    /* 80B270E0 */ void Draw();
-    /* 80B27174 */ void createHeapCallBack(fopAc_ac_c*);
-    /* 80B27194 */ void ctrlJointCallBack(J3DJoint*, int);
-    /* 80B271EC */ void getType();
-    /* 80B27260 */ void isDelete();
+    /* 80B268F8 */ int create();
+    /* 80B26BE4 */ int CreateHeap();
+    /* 80B2708C */ int Delete();
+    /* 80B270C0 */ int Execute();
+    /* 80B270E0 */ int Draw();
+    /* 80B27174 */ static int createHeapCallBack(fopAc_ac_c*);
+    /* 80B27194 */ static int ctrlJointCallBack(J3DJoint*, int);
+    /* 80B271EC */ u8 getType();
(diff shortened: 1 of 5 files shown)


include/d/actor/d_a_npc_uri.h
```diff
@@ -11,80 +11,179 @@
...
+    /* 0x1010 */ u8 field_0x1010;
+    /* 0x1011 */ u8 field_0x1011;
+    /* 0x1012 */ u8 field_0x1012;
+    /* 0x1014 */ int field_0x1014;
 };
 
 STATIC_ASSERT(sizeof(daNpc_Uri_c) == 0x1018);
 
+struct daNpc_Uri_HIOParam {
+    /* 0x00 */ f32 field_0x00;  // 200.0f
```
TakaRikka, line 138: instead of putting the float values as comments here, i'd rather put a comment in the actual data initializer with the name of each value. reasoning being the values are supposed to be tweaked and such, while the names arent

src/d/actor/d_a_npc_uri.cpp
```diff
@@ -1,1905 +1,1844 @@
...
+void daNpc_Uri_c::afterJntAnm(int param_1) {
+    if (param_1 == 1) {
+        mDoMtx_stack_c::YrotM(mStagger.getAngleZ(1));
+        mDoMtx_stack_c::ZrotM(-mStagger.getAngleX(1));
+    } else if (param_1 == 4) {
+        mDoMtx_stack_c::YrotM(mStagger.getAngleZ(0));
+        mDoMtx_stack_c::ZrotM(mStagger.getAngleX(0));
+    }
+
+    if ((f32)0.0 < field_0x1004) {
```
TakaRikka, line 451: why the cast instead of 0.0f?

src/d/actor/d_a_npc_uri.cpp
```diff
@@ -1,1905 +1,1844 @@
...
+        return TRUE;
+    }
+    return FALSE;
 }
 
 /* 80B288DC-80B28910 00221C 0034+00 1/1 0/0 0/0 .text chkPlayerGetWoodShield__11daNpc_Uri_cFv */
-void daNpc_Uri_c::chkPlayerGetWoodShield() {
-    // NONMATCHING
+BOOL daNpc_Uri_c::chkPlayerGetWoodShield() {
+    return g_dComIfG_gameInfo.info.getPlayer().getCollect().isCollect(2, 0);
```
TakaRikka, line 913: should use `dComIfGs_isCollectShield` in here

src/d/actor/d_a_npc_uri.cpp
```diff
@@ -1,1905 +1,1844 @@
...
+        break;
+    case 1:
+        uVar8 = 1;
+        break;
+    case 2:
+        int local_48;
+        if (talkProc(NULL, 0, NULL, 0) && mFlow.checkEndFlow() &&
+            (s32)mFlow.getEventId(&local_48) == 1)
+        {
+            if (mItemId == -1) {
```
TakaRikka, line 1121: since it's an actor ID, you can use `fpcM_ERROR_PROCESS_ID_e`

src/d/actor/d_a_npc_uri.cpp
```diff
@@ -1919,20 +1858,20 @@ s32 daNpc_Uri_c::getNeckJointNo() {
...
+    return param_0 == 4;
 }
 
 /* 80B2CD74-80B2CD84 0066B4 0010+00 1/0 0/0 0/0 .text            checkRemoveJoint__11daNpc_Uri_cFi
  */
-void daNpc_Uri_c::checkRemoveJoint(int param_0) {
-    // NONMATCHING
+BOOL daNpc_Uri_c::checkRemoveJoint(int param_0) {
+    return param_0 == 8;
 }
```
TakaRikka, line 1876: with these small funcs being all the way here at the end of the TU, i have a strong suspicion they are supposed to be inlined weak functions

hatal175 (approved):

PR #2394, opened 2025-04-14
d_a_e_ms work

carter-ktb21:

PR #2405, opened 2025-04-21
d_a_e_ms Equivalent

ItsNiklas:
Builds on top of #2394, to fix all outstanding issues. Except for weak function order (I think) regarding `setMidnaBindEffect`, which gets placed too early.

 Some cleaning (clang format?!), but mostly fixing regallocs etc. ~Also a nasty goto fakematch in `search_dokuro`...~


hatal175:
This matches without the goto https://decomp.me/scratch/LkReJ

hatal175 (approved):

PR #2407, opened 2025-04-22
d_menu_letter equivalent + various debug changes

hatal175:


include/dolphin/gx/GXGeometry.h
```diff
@@ -30,7 +30,7 @@ static inline void GXEnd(void) {
     extern GXBool __GXinBegin;
     extern void OSPanic(const char* file, int line, const char* msg, ...);
     if (!__GXinBegin) {
-        OSPanic(__FILE__, 118, "GXEnd: called without a GXBegin");
+        OSPanic(__FILE__, 127, "GXEnd: called without a GXBegin");
```
TakaRikka, line 33: if this and the other dolphin SDK changes are for debug, i'd prefer they be reverted for now, because Shield is actually supposed to use the wii SDK and i want to keep the dolphin SDK synced with dolsdk2004

TakaRikka (approved):

PR #2411, opened 2025-04-26
d_a_npc_bou Equivalent

YunataSavior:


include/d/actor/d_a_npc_bou.h
```diff
@@ -176,18 +188,55 @@ class daNpc_Bou_c : public daNpcT_c {
...
+    /* 0xFE8 */ u8 field_0xfe8[0xFF4 - 0xFE8];
+    /* 0xFF4 */ u8 field_0xff4;
+    /* 0xFF5 */ u8 field_0xff5[0xFF8 - 0xFF5];
+    /* 0xFF8 */ u8 field_0xff8;
 };
 
 STATIC_ASSERT(sizeof(daNpc_Bou_c) == 0xffc);
 
+/* 809727B4-809727BC 005914 0008+00 1/0 0/0 0/0 .text getEyeballMaterialNo__11daNpc_Bou_cFv */
+s32 daNpc_Bou_c::getEyeballMaterialNo() {
```
hatal175, line 209: Didn't work as inline?

YunataSavior: This was done for weak ordering purposes. When they're declared inline, they'll still show up as separate functions, but their definition in the assembly will show up directly after the function that calls it.

hatal175 (approved):

PR #2419, opened 2025-05-04
d_a_cstatue Matching

soruh:


soruh:
It seems the condition on `checkNotSmallType` in `procCopyRodSwingInit` was  de Morganed incorrectly and the inline was thus also inverted. So fixing the inline broke `procCopyRodSwingInit`.

include/d/actor/d_a_alink.h
```diff
@@ -3545,6 +3615,12 @@ class daAlink_c : public daPy_py_c {
     bool checkSpinnerRideWait() const {
         return mProcID == PROC_SPINNER_WAIT && mProcVar2.field_0x300c == 0;
     }
+    bool checkCopyRodSwingMode() { return mProcID == PROC_COPY_ROD_SWING; }
+    bool checkCopyRodSwingModeInit() {
+        return mProcID == PROC_COPY_ROD_SWING && mProcVar2.field_0x300c != 0;
+    }
+    double transAnimeMoveX() { return field_0x3464; }
+    double transAnimeMoveZ() { return field_0x3468; }
```
TakaRikka, line 3623: both should return float

src/d/actor/d_a_cstatue.cpp
```diff
@@ -1,1640 +1,1286 @@
...
+    cs_b_fall = 7,       //	Copy Statue - B - Fall
+    cs_b_jump = 8,       //	Copy Statue - B - Jump
+    cs_b_walk = 9,       //	Copy Statue - B - Walk
+    cs_fall = 10,        //	Copy Statue - Fall
+    cs_start = 11,       //	Copy Statue - Start
+    cs_start_demo = 12,  //	Copy Statue - Start - Demo
+    cs_stop = 13,        //	Copy Statue - Stop
+    cs_swing = 14,       //	Copy Statue - Swing
+    cs_swingre = 15,     //	Copy Statue - Swing Re
+    cs_walk = 16,        //	Copy Statue - Walk
```
TakaRikka, line 35: not a major thing but i prefer these all uppercase

src/d/actor/d_a_cstatue.cpp
```diff
@@ -1,1640 +1,1286 @@
...
-
-/* 806647D8-80664820 001218 0048+00 1/0 0/0 0/0 .text            __dt__12J3DFrameCtrlFv */
-// J3DFrameCtrl::~J3DFrameCtrl() {
-extern "C" void __dt__12J3DFrameCtrlFv() {
-    // NONMATCHING
+int daCstatue_c::create() {
+    if (!fopAcM_CheckCondition(this, 8)) {
+        new (this) daCstatue_c();
+        fopAcM_OnCondition(this, 8);
+    }
```
TakaRikka, line 227: you can use the fopAcM_SetupActor macro for this

src/d/actor/d_a_cstatue.cpp
```diff
@@ -1,1640 +1,1286 @@
...
-/* ############################################################################################## */
 /* 806677F4-80667800 00015C 000C+00 0/1 0/0 0/0 .rodata          normalLocalBallPos$4911 */
-#pragma push
-#pragma force_active on
-SECTION_RODATA static u8 const normalLocalBallPos[12] = {
-    0x42, 0x8C, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-};
-COMPILER_STRIP_GATE(0x806677F4, &normalLocalBallPos);
-#pragma pop
+static Vec const normalLocalBallPos = {70.0f, 0.0f, 0.0f};
```
TakaRikka, line 454: all data that ends with a $ and numbers like `normalLocalBallPos$4911` means it's in-function static data, so these can be moved inside the function they're used in

src/d/actor/d_a_cstatue.cpp
```diff
@@ -1,1640 +1,1286 @@
...
+                } else {
+                    camera->mCamera.Start();
+                    dComIfGp_event_reset();
+                }
+            }
+        }
+    } else {
+        fopAcM_orderPotentialEvent(this, 2, (u16)-1, 0);
+        eventInfo.onCondition(2);
+        return;
```
TakaRikka, line 902: useless return keyword can be removed

hatal175 (approved):

PR #2421, opened 2025-05-05
d_a_e_sg Matching

### LINKED ISSUE #594: d_a_e_sg.cpp

ItsNiklas:
- `e_sg_class`
- `fopEn_enemy_c`
- `15.20 kB (0.12%)`

Skullfish from Lakebed Temple.

Overall happy with this class considering the bitmask joint shenanigans that are going on. There are some more weird sections, but it works for now.
Also, I had to fill some fishing rod fields and 'kbox' fields which seems to be the other target option.

---

Resolves #594.


src/d/actor/d_a_e_sg.cpp
```diff
@@ -1,1256 +1,1321 @@
...
 };
 
 /* 8078A22C-8078A26C 0000EC 0040+00 1/1 0/0 0/0 .text            __ct__12daE_SG_HIO_cFv */
 daE_SG_HIO_c::daE_SG_HIO_c() {
-    // NONMATCHING
+    this->mUnk0 = -1;
+    this->mScaleFactor = 1.2f;
+    this->mMovementSpeed = 10.0f;
+    this->mSearchSpeed = 13.0f;
+    this->mPlayerHeightThreshold = 500.0f;
```
TakaRikka, line 55: `this->` can be removed

src/d/actor/d_a_e_sg.cpp
```diff
@@ -1,1256 +1,1321 @@
...
+
+        sVar1 -= actor->current.angle.y;
+        fVar1 = sVar1 * 5.0f;
+        if (fVar1 > 4000.0f) {
+            fVar1 = 4000.0f;
+        } else {
+            if (fVar1 < -4000.0f) {
+                fVar1 = -4000.0f;
+            }
+        }
```
TakaRikka, line 530: else if

src/d/actor/d_a_e_sg.cpp
```diff
@@ -1,1256 +1,1321 @@
...
+
+    e_sg_class* a_this = static_cast<e_sg_class*>(i_this);
+    fopAcM_SetupActor(i_this, e_sg_class);
+
+    cPhs__Step step = (cPhs__Step)dComIfG_resLoad(&a_this->mPhaseReq, "E_sg");
+
+    if (step == cPhs_COMPLEATE_e) {
+        if (!fopAcM_entrySolidHeap(i_this, useHeapInit, 4000)) {
+            return cPhs_ERROR_e;
+        } else {
```
TakaRikka, line 1218: else block can be removed since theres an early return

src/d/actor/d_a_e_sg.cpp
```diff
@@ -1,1256 +1,1321 @@
...
+        i_this->mGroundY = fVar71;
+        i_this->home.pos.y = i_this->mGroundY - 250.0f + cM_rndFX(50.0f);
+    }
+    i_this->mPlayerAngle = fopAcM_searchPlayerAngleY(i_this);
+    i_this->mPlayerDist = fopAcM_searchPlayerDistance(i_this);
+    i_this->mTargetDist = (i_this->home.pos - player->current.pos).abs();
+
+    for (int i = 0; i < 3; i++) {
+        if (i_this->mCooldowns[i] != 0) {
+            i_this->mCooldowns[i]--;
```
TakaRikka, line 1045: i think i'd prefer if this was called `mTimers` instead

PR #2429, opened 2025-05-08
d_a_npc_cd work

carter-ktb21:

PR #2433, opened 2025-05-10
d_a_npc_cd Equivalent

ItsNiklas:
Another janitor job from #2429 🧹 (Some random changes are to try and match debug)

Sad that this does not match, the only two issues I can see are that `daNpcCd_HIO_c::~daNpcCd_HIO_c()` gets placed too early and that the `daNpcCd_c::setAttention(int)::a_eyeOfsTbl guard` is placed in `.sbss`, instead of being "extern"? Not sure if that is an easy fix or not a problem?


src/d/actor/d_a_npc_cd.cpp
```diff
@@ -238,29 +250,12 @@ int daNpcCd_c::NpcCreate(int param_1) {
...
-    if (mpMorf != NULL && mpMorf->getModel() == NULL) {
-        mpMorf->stopZelAnime();
-        mpMorf = NULL;
-    }
-
-    if (mpMorf == NULL) {
-        return 0;
-    }
+    // #if VERSION == VERSION_SHIELD_DEBUG
+    // #endif
```
TakaRikka, line 254: can these comments be removed?

ItsNiklas: Yes

TakaRikka (approved):

PR #2435, opened 2025-05-12
d_a_np_seid equivalent (minus rodata oddity)

carter-ktb21:


hatal175 (commented):
Mark as equivalent in configuration file (preferably with reason)

TakaRikka (approved):

PR #2437, opened 2025-05-13
d_a_obj_ganonwall2 Matching

Hyenate:


hatal175 (commented):
Good Match. If you're getting into too much issues matching this while making the changes, just say so, and someone will get to it at some point.

Hyenate:
I'll be sure to fix what I can as soon as possible, and I'll definitely ask for help if I'm having any trouble. I referenced d_a_obj_ganonwall for some of the matching, should I see if I can make some of the same fixes there?

hatal175:
If you want to. You can always reference the debug rom.

hatal175 (approved):

PR #2438, opened 2025-05-14
d_a_npc_fairy_seirei Equivalent

Stbn-antier:
Solves #757 
100% matching, still only Equivalent


Stbn-antier:
Thanks for the reviews as always @hatal175 !

hatal175 (approved):

PR #2441, opened 2025-05-15
Work on d_a_e_mk - beeg monkey :)

carter-ktb21:

PR #2443, opened 2025-05-16
d_a_e_mk Work

ItsNiklas:
Fix some issues remaining with #2441. Remainder is probably some combination of casting to the base actor class to fix regallocs, which I cannot be bothered to do.

To be rebased.


ItsNiklas:
Closing since changes are now in #2446..

PR #2445, opened 2025-05-16
d_a_obj_nameplate matching

ljw947:
Solves https://github.com/zeldaret/tp/issues/1099 matching

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -1571,7 +1571,7 @@ config.libs = [
     ActorRel(NonMatching, "d_a_obj_magne_arm"),
     ActorRel(MatchingFor("GZ2E01"), "d_a_obj_metalbox"),
     ActorRel(MatchingFor("GZ2E01"), "d_a_obj_mgate"),
-    ActorRel(Equivalent, "d_a_obj_nameplate"),
+    ActorRel(Matching, "d_a_obj_nameplate"),
     ActorRel(NonMatching, "d_a_obj_ornament_cloth"),
     ActorRel(NonMatching, "d_a_obj_rope_bridge"),
     ActorRel(MatchingFor("GZ2E01"), "d_a_obj_sWallShutter"),
--- a/configure.py
+++ b/configure.py
@@ -1571,7 +1571,7 @@ config.libs = [
     ActorRel(NonMatching, "d_a_obj_magne_arm"),
     ActorRel(MatchingFor("GZ2E01"), "d_a_obj_metalbox"),
     ActorRel(MatchingFor("GZ2E01"), "d_a_obj_mgate"),
-    ActorRel(NonMatching, "d_a_obj_nameplate"),
+    ActorRel(Equivalent, "d_a_obj_nameplate"),
     ActorRel(NonMatching, "d_a_obj_ornament_cloth"),
     ActorRel(NonMatching, "d_a_obj_rope_bridge"),
     ActorRel(MatchingFor("GZ2E01"), "d_a_obj_sWallShutter"),
--- a/include/d/actor/d_a_obj_nameplate.h
+++ b/include/d/actor/d_a_obj_nameplate.h
@@ -1,8 +1,9 @@
 #ifndef D_A_OBJ_NAMEPLATE_H
 #define D_A_OBJ_NAMEPLATE_H
 
-#include "f_op/f_op_actor_mng.h"
 #include "d/d_cc_d.h"
+#include "d/d_com_inf_game.h"
+#include "d/d_msg_flow.h"
 
 /**
  * @ingroup actors-objects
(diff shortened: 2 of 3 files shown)


configure.py
```diff
@@ -1571,7 +1571,7 @@ def MatchingFor(*versions) -> bool:
     ActorRel(NonMatching, "d_a_obj_magne_arm"),
     ActorRel(MatchingFor("GZ2E01"), "d_a_obj_metalbox"),
     ActorRel(MatchingFor("GZ2E01"), "d_a_obj_mgate"),
-    ActorRel(NonMatching, "d_a_obj_nameplate"),
+    ActorRel(Equivalent, "d_a_obj_nameplate"),
```
ItsNiklas, line 1574: It is actually not Equivalent, but we can bring it to Matching!

include/d/actor/d_a_obj_nameplate.h
```diff
@@ -21,21 +22,54 @@ class daObjNameplate_c : public fopAc_ac_c {
     /* 805942B8 */ void setCcSphere();
     /* 8059434C */ void calcAngle();
     /* 805945CC */ void calcSpring();
-    /* 80594754 */ void checkHitAngle();
+    /* 80594754 */ int checkHitAngle();
     /* 805947F8 */ void messageProc();
 
-    static u8 const M_attr[32];
+    ~daObjNameplate_c() { dComIfG_resDelete(&this->mPhase, "J_Hyosatu"); }
```
ItsNiklas, line 28: We need to use l_arcName here, so lets create the static variable inside the class

`static char* l_arcName;`

And define it somewhere in the cpp file `char* daObjNameplate_c::l_arcName = "J_Hyosatu";`

src/d/actor/d_a_obj_nameplate.cpp
```diff
@@ -1,378 +1,362 @@
...
-    // NONMATCHING
+    switch (mShowMessage) {
+    case 0:
+        if (mMessageFlowId != -1 &&
+            cLib_distanceAngleS(shape_angle.GetY(), fopAcM_searchPlayerAngleY(this)) < 0x3000)
+        {
+            fopAcM_OnStatus(this, 0);
+            cLib_onBit<u32>(attention_info.flags, 0x4000000A);
+            attention_info.distances[fopAc_attn_TALK_e] = 0x15;
+            attention_info.distances[fopAc_attn_BATTLE_e] = 0x15;
```
ItsNiklas, line 226:
```suggestion
            attention_info.distances[fopAc_attn_SPEAK_e] = 0x15;
```

src/d/actor/d_a_obj_nameplate.cpp
```diff
@@ -1,378 +1,362 @@
...
 /* 80594F28-80594F2C -00001 0004+00 3/3 0/0 0/0 .data            l_arcName */
-SECTION_DATA static void* l_arcName = (void*)&d_a_obj_nameplate__stringBase0;
+static char* l_arcName = "J_Hyosatu";
 
-/* 80594950-805949C8 000930 0078+00 1/1 0/0 0/0 .text            createSolidHeap__FP10fopAc_ac_c */
-static void createSolidHeap(fopAc_ac_c* param_0) {
-    // NONMATCHING
-}
+int daObjNameplate_c::createHeap() {
+    J3DModelData* modelData = (J3DModelData*)dComIfG_getObjectRes(l_arcName, "J_Hyosatu.bmd");
```
ItsNiklas, line 257: There is actually a different string here... 
```suggestion
    J3DModelData* modelData = (J3DModelData*)dComIfG_getObjectRes(l_arcName, "J_Hyousatu.bmd");
```

ItsNiklas (commented):
Thank you for the contribution! Just a few typos to fix, so this is `Matching`! A few more style changes to go, but thank you 😃

src/d/actor/d_a_obj_nameplate.cpp
```diff
@@ -1,378 +1,362 @@
...
+    VECAdd(&attention_info.position, &local_18, &attention_info.position);
+
+    eyePos.y = eyePos.y - 120.0f;
+
+    mMessageFlowId = fopAcM_GetParam(this);
+
+    initBaseMtx();
+    initCcSphere();
+
+    return;
```
hatal175, line 46: No need for return at the end of the function

src/d/actor/d_a_obj_nameplate.cpp
```diff
@@ -1,378 +1,362 @@
...
+        if (play.getEvent().runCheck() && eventInfo.checkCommandTalk() != 0 &&
+            mMsgFlow.doFlow(this, NULL, 0) != 0)
+        {
+            play.getEvent().reset();
+            mShowMessage = 0;
+        }
+        break;
+    }
+
+    return;
```
hatal175, line 250: No need for return here.

configure.py
ljw947: think I got confused what equivalent vs matching meant 😃

ljw947:
@ItsNiklas @hatal175 comments addressed

include/d/actor/d_a_obj_nameplate.h
```diff
@@ -21,21 +22,56 @@ class daObjNameplate_c : public fopAc_ac_c {
...
     /* 8059434C */ void calcAngle();
     /* 805945CC */ void calcSpring();
-    /* 80594754 */ void checkHitAngle();
+    /* 80594754 */ int checkHitAngle();
     /* 805947F8 */ void messageProc();
 
-    static u8 const M_attr[32];
+    struct M_attrs {
+        /* 0x00 */ f32 radius1;
+        /* 0x04 */ f32 float_field_0x04;
```
hatal175, line 30: One last thing I haven't noticed before. The code standard is just field_0x04 not short_field_0x04 or float_field_0x04.
Could you just rename both here and in the class itself?

A quick search replace just do.

hatal175 (approved):

PR #2453, opened 2025-05-22
Update dtk-template

LagoLunatic:
* Progress upload to frogress has been removed.
* sjiswrap updated to support .inc files, so d_event_debug can be fixed now.
* objdiff 3 changed its algorithm so the bot will report a lot of fake diffs for this PR.

Note that I also enabled `functionRelocDiffs=data_value` in the progress report config, which allows for more types of regressions to be caught by diffing progress reports which would currently slip through in some cases that don't affect codegen, such as calling the wrong function symbol or using the wrong float literal. This is optional, can be disabled by commenting that line out if it causes any unwanted behavior.


TakaRikka:
are there any important regressions to note? matched code dropping 23084 bytes seems like a lot

LagoLunatic:
> are there any important regressions to note? matched code dropping 23084 bytes seems like a lot

I fixed most of the 100% regressions, it was pretty much symbols and splits being wrong because they're missing from TP's maps. Also linked the TUs now that the splits are fixed.

I don't understand what the last three 100% regressions are (those -512 bytes):
```
$ ninja changes | grep 100\\.00%
                                                                        framework/d/actor/d_a_npc_cd | matched_data | 100.00% ->  99.97%
                                                                        framework/dolphin/os/__start | fuzzy_match  | 100.00% ->  99.84%
                                                                        framework/dolphin/os/__start | matched_code | 100.00% ->  36.92%
                                                                                             __start | fuzzy_match  | 100.00% ->  99.66%
                                                                                         __init_data | fuzzy_match  | 100.00% ->  99.90%
```

They show seem to 100% in objdiff GUI, so I'm not sure how objdiff-cli is counting them.

TakaRikka (approved):

PR #2512, opened 2025-06-30
some more wii tests/improvements

TakaRikka:
wii appears to use `-ipa file` which solves most inline issues, but not all. additionally, using a patched GC 3.0 compiler that reverts the changes that disallow character constants greater than 4 bytes. we're likely missing the exact original compiler used


include/Z2AudioLib/Z2AudioCS.h
```diff
@@ -0,0 +1,22 @@
+#ifndef Z2AUDIOCS_H
+#define Z2AUDIOCS_H
+
+#include <revolution/wpad.h>
```
hatal175, line 4: Note that my previous commit has a a very short stub for this file in a different location. Could you choose one and adjust include in Z2LinkMgr.h? Also startLevel is static.

PR #2517, opened 2025-07-05
d_a_e_po 99% Matching

Stbn-antier:
Only one method is not matching (e_po_dead).

・Most variables and members still have default names, I didn't have time to understand the logic
・A lot of smelly code, like a data padding at the beginning to align .bss section
・Also don't know how to properly find or name all the joints for the model

Please comment if formatting is wrong for anims and stuff.
If someone wants to finish this TU please do, I won't have time to contribute for a bit longer ^^'


src/d/actor/d_a_e_po.cpp
```diff
@@ -2703,538 +2825,366 @@ static void mPo_tgHitCallback(fopAc_ac_c* param_0, dCcD_GObjInf* param_1, fopAc_
...
+        a_this->BitSW3 = fopAcM_GetParam(i_this) >> 0x18;
+        if (ZREG_S(0) != 0) {
+            a_this->BitSW = ZREG_S(0);
+        }
+        if (mArg0Check(a_this, 3) == 0 && mArg0Check(a_this, 0xFF) == 2) {
+            a_this->BitSW2 = 0xFF;
+        }
+
+        OS_REPORT("\n\n\n");
+        OS_REPORT("E_PO：i_this->arg0 %d\n", a_this->mArg0);
```
Luzagar: it should use i_this instead of a_this

src/d/actor/d_a_e_po.cpp
```diff
@@ -1,2698 +1,2820 @@
...
+    mFallChanceTime = 50;
+    mAttackChanceTime = 50;
+    mModelSize = 1.0f;
 }
 
-/* 8074C5EC-8074C660 00018C 0074+00 13/13 0/0 0/0 .text            mArg0Check__FP10e_po_classs */
-static void mArg0Check(e_po_class* param_0, s16 param_1) {
-    // NONMATCHING
+/* 80757E5C-80757E98 000054 003C+00 9/11 0/0 0/0 .bss             l_HIO */
+static u8 lbl_221_bss_45;
```
Luzagar: should be renamed l_HIOInit

src/d/actor/d_a_e_po.cpp
```diff
@@ -2703,538 +2825,366 @@ static void mPo_tgHitCallback(fopAc_ac_c* param_0, dCcD_GObjInf* param_1, fopAc_
...
+        }
+        if (mArg0Check(a_this, 3) == 0 && mArg0Check(a_this, 0xFF) == 2) {
+            a_this->BitSW2 = 0xFF;
+        }
+
+        OS_REPORT("\n\n\n");
+        OS_REPORT("E_PO：i_this->arg0 %d\n", a_this->mArg0);
+        OS_REPORT("E_PO：i_this->bitsw %d\n", a_this->BitSW);
+        OS_REPORT("E_PO：i_this->bitsw2 %d\n", a_this->BitSW2);
+        OS_REPORT("E_PO：i_this->bitsw3 %d\n", a_this->BitSW3);
```
Luzagar: might be not accurate but maybe keep the member name style ? ex bitsw instended of BitSW

Luzagar (changes requested):

include/d/actor/d_a_e_po.h
```diff
@@ -1,27 +1,179 @@
...
+    /* 0xE54 */ dCcU_AtInfo mAtInfo;
+    /* 0xE78 */ u32 mParticleKey6;
+    /* 0xE7C */ u32 mParticleKey5[2];
+    /* 0xE84 */ u32 mParticleKey[4];
+    /* 0xE94 */ u32 mParticleKey2[4];
+    /* 0xEA4 */ u32 mParticleKey3[2];
+    /* 0xEAC */ u32 mParticleKey4[2];
+    /* 0xEB4 */ dPa_hermiteEcallBack_c field_0xEB4;
+    /* 0xECC */ u8 field_0xECC;
+    /* 0xECD */ u8 field_0xECD[0xee0 - 0xECD];  // Padding
```
hatal175, line 142: I think this padding can be removed

include/d/actor/d_a_e_po.h
```diff
@@ -1,27 +1,179 @@
...
 };
 
 STATIC_ASSERT(sizeof(e_po_class) == 0xee0);
 
 class daE_PO_HIO_c {
+public:
     /* 8074C54C */ daE_PO_HIO_c();
-    /* 80756DE8 */ ~daE_PO_HIO_c();
+    /* 80756DE8 */ virtual ~daE_PO_HIO_c() {};
```
hatal175: Remove ;

include/d/actor/d_a_e_po.h
```diff
@@ -1,27 +1,179 @@
...
+    /* 0x2c */ s16 mReviveTime2;
+    /* 0x2e */ s16 mReviveTime3;
+    /* 0x30 */ s16 mReviveTime4;
+    /* 0x32 */ s16 mFallChanceTime;
+    /* 0x34 */ s16 mAttackChanceTime;
+    /* 0x36 */ s16 field_0x36;
+    /* 0x38 */ f32 mModelSize;
 };
 
+inline void dComIfGs_addPohSpiritNum() {
```
hatal175: Put with other dcom functions

include/d/d_save.h
```diff
@@ -394,7 +394,12 @@ class dSv_player_collect_c {
     void offCollectMirror(u8 i_item);
     BOOL isCollectMirror(u8 i_item) const;
 
-    u8 getPohNum() const { return mPohNum; }
+    u8 getPohNum() { return mPohNum; }
```
hatal175: Please return this to const, I believe this is what causes the broken match

src/d/actor/d_a_e_po.cpp
```diff
@@ -1,2698 +1,2820 @@
...
+                !dComIfGp_getAttention()->LockonTruth() ||
+                dComIfGp_getAttention()->GetLockonList(0)->getActor() != a_this)
+            {
+                break;
+            }
+        }
+        i_this->field_0x5C0 = true;
+        i_this->mSound1.startCreatureVoice(Z2SE_EN_PO_V_FIND, -1);
+        if (!a_this->eventInfo.checkCommandDemoAccrpt()) {
+            fopAcM_orderPotentialEvent((fopAc_ac_c*)i_this, 2, -1, 0);
```
hatal175: I see (fopAc_ac_c*)i_this in a lot of places. Would using a_this screw up the match?

Stbn-antier: It did before but after the cleanup it seems I can put a_this while keeping the match

include/d/actor/d_a_e_po.h
Stbn-antier: It causes a failed assertion on my side if I remove it

src/d/actor/d_a_e_po.cpp
Stbn-antier: It causes a mismatch if I do, when the variable is used the assembly tells me:
```
lis r3, lbl_221_bss_45@ha  
stb r0, lbl_221_bss_45@l(r3)
```
Should I still name it l_HIOInit to keep it coherent ?

src/d/actor/d_a_e_po.cpp
Stbn-antier: Other OS_REPORT have it written as either bitSw (b_dr) or BitSW (npc_saru, npc_taro), and the latter is used for inline names, so I think BitSW may be the convention used by the devs for this kind of variable.
As it would be the only instance of "bitsw" in the code, I was thinking it's a bit more coherent as "BitSW"

include/d/actor/d_a_e_po.h
hatal175: Oh yeah, sorry, missed that it was ee0 not ed0

src/d/actor/d_a_e_po.cpp
hatal175: Search the symbols.txt for lbl_221_bss_45 and rename it there as well

PR #2523, opened 2025-07-08
A means of de-inlining the daNpcT_c ctors

YunataSavior:
So IDK how legit this is, but I can get the daNpcF_c ctor functions to non-inline via the following pragma in configure.py:
```
ActorRel(NonMatching, "d_a_npc_ash", extra_cflags=['-pragma "inline_max_size(494)"']),
```

Unfortunately, using this inline setting as-so for all the rels causes regressions, so I can only do this on a per-TU basis.

**I'm submitting this as a draft just as a demonstration. Let's discuss the ramifications and if there might be other means of de-inlining.**


configure.py
```diff
@@ -412,6 +412,10 @@ def MatchingFor(*versions) -> bool:
     return config.version in versions
 
 
+# For whatever reason, daNpcF_c::daNpcF_c gets inlined.
+#  Using the following prevents that from happening:
```
TakaRikka, line 416: i would just add a TODO that this should be looked into more extensively in the future and is just a temporary hack

PR #2525, opened 2025-07-09
d_a_demo00 work

carter-ktb21:
This was another really complicated class, so it's a bit messy.

Hopefully, however, it will serve as a decent foundation to be cleaned up.

### DIFF
--- a/include/d/actor/d_a_demo00.h
+++ b/include/d/actor/d_a_demo00.h
@@ -8,11 +8,25 @@ struct demo_s1_ke_s {
 public:
     /* 804A4420 */ ~demo_s1_ke_s();
     /* 804A86B4 */ demo_s1_ke_s();
+
+    /* 0x000 */ cXyz field_0x0[16];
+    /* 0x0C0 */ cXyz field_0xc0[16];
+    /* 0x180 */ cXyz field_0x180;
 };
 
 class daDemo00_resID_c {
 public:
     /* 804A430C */ void reset();
+
+    /* 0x00 */ int field_0x0;
+    /* 0x04 */ int field_0x4;
+    /* 0x08 */ int field_0x8;
+    /* 0x0C */ int field_0xc;
+    /* 0x10 */ int field_0x10;
+    /* 0x14 */ int field_0x14;
+    /* 0x18 */ int field_0x18;
+    /* 0x1C */ int field_0x1c;
+    /* 0x20 */ int field_0x20;
 };
 
 class daDemo00_model_c {
@@ -33,7 +47,7 @@ public:
     /* 804A4388 */ ~daDemo00_c();
     /* 804A45A0 */ void setBaseMtx();
     /* 804A4948 */ void setShadowSize();
-    /* 804A4F74 */ void createHeap();
+    /* 804A4F74 */ int createHeap();
     /* 804A5798 */ void actStandby(dDemo_actor_c*);
     /* 804A594C */ void actPerformance(dDemo_actor_c*);
     /* 804A604C */ void actLeaving(dDemo_actor_c*);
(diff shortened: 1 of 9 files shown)


include/d/d_demo.h
```diff
@@ -343,6 +346,8 @@ class jstudio_tAdaptor_message : public JStudio::TAdaptor_message {
 
 };  // namespace
 
+inline static int dDemo_setDemoData(fopAc_ac_c*, u8, mDoExt_McaMorf*, char const*, int, u16*, u32, s8);
```
hatal175: Why is this declared as inline static in header file? it's not static nor inline as far as I can see and is what probably causes the regression

include/d/actor/d_a_demo00.h
```diff
@@ -30,18 +61,85 @@ class daDemo00_model_c {
...
+    /* 0x06B1 */ s8 field_0x6b1;
+    /* 0x06B2 */ u8 field_0x6b2;
+    /* 0x06B3 */ s8 field_0x6b3;
+    /* 0x06B4 */ s8 field_0x6b4;
+    /* 0x06B5 */ u8 field_0x6b5;
+    /* 0x06B6 */ u8 field_0x6b6;
+    /* 0x06B7 */ u8 field_0x6b7;
+    /* 0x06B8 */ u8 field_0x6b8;
+    /* 0x06B9 */ s8 field_0x6b9;
+    /* 0x06BA */ u8 field_0x6ba[0x6bc - 0x6ba];
```
hatal175: I think field_0x6ba can be removed

src/d/actor/d_a_demo00.cpp
```diff
@@ -4,1505 +4,1897 @@
...
-/* 804A5750-804A5798 001530 0048+00 1/0 0/0 0/0 .text            __dt__12J3DFrameCtrlFv */
-// J3DFrameCtrl::~J3DFrameCtrl() {
-extern "C" void __dt__12J3DFrameCtrlFv() {
+int daDemo00_c::createHeap() {
     // NONMATCHING
+    if (mModel.mID.mShapeID != -1) {
+        J3DModelData* modelData = (J3DModelData*)dComIfG_getObjectIDRes(dStage_roomControl_c::getDemoArcName(), (u16)mModel.mID.mShapeID);
+        mDoExt_bckAnmRemove(modelData);
+
+        #ifdef DEBUG
```
hatal175: I don't think you need the ifdef - it should optimize the if away since everything inside it is defined as nothing

src/d/actor/d_a_demo00.cpp
```diff
@@ -4,1505 +4,1897 @@
...
+        if (mBgc != NULL) {
+            cXyz sp70(current.pos.x, current.pos.y + 100.0f, current.pos.z);
+            mBgc->mGndChk.SetPos(&sp70);
+            mBgc->field_0x54 = dComIfG_Bgsp().GroundCross(&mBgc->mGndChk);
+            field_0x6a1 = 1;
+        }
+
+        setBaseMtx();
+
+        if (actor->checkEnable(0x40)) {
```
hatal175: Use ENABLE_ANM_FRAME_e

src/d/actor/d_a_demo00.cpp
```diff
@@ -4,1505 +4,1897 @@
...
-#pragma force_active on
-SECTION_RODATA static f32 const lit_5286 = 32768.0f;
-COMPILER_STRIP_GATE(0x804A88EC, &lit_5286);
-#pragma pop
-
-/* 804A61F0-804A6428 001FD0 0238+00 1/1 0/0 0/0 .text            teduna_calc__FP4cXyzP4cXyzP4cXyzsi
- */
-static void teduna_calc(cXyz* param_0, cXyz* param_1, cXyz* param_2, s16 param_3, int param_4) {
+/* 804A88CC-804A88D0 00006C 0004+00 1/1 0/0 0/0 .rodata          l_itemNo$localstatic3$execute__10daDemo00_cFv */
+// static u8 const l_itemNo[4] = {
```
hatal175: Comment can be removed

src/d/actor/d_a_demo00.cpp
```diff
@@ -4,1505 +4,1897 @@
...
+        sp7c.set(0.0f, -30.0f, 60.0f);
+    }
+
+    mDoMtx_stack_c::multVec(&sp7c, &sp88);
+
+    for (int i = 0; i < 16; i++) {
+        *param_3 = *param_1 - (sp70 * (i / 15.0f));
+        f32 fVar2 = cM_ssin((i / 15.0f) * 32768.0f);
+        *param_3 += sp88 * fVar2;
+        param_3->y += fVar2 * (fVar1 * cM_ssin(g_Counter.mCounter0 * 0x9C4 + i * 0x640));
```
hatal175: Use 2500 instead of 0x9C4 and 1600 instead of 0x640

src/d/actor/d_a_demo00.cpp
```diff
@@ -4,1505 +4,1897 @@
...
+    } else if (field_0x5d4 != NULL) {
+        if (mpModelMorf != NULL && mModel.mID.field_0x4 != mModel.field_0x0.field_0x4) {
+            J3DAnmTransform* anm = (J3DAnmTransform*)dComIfG_getObjectIDRes(dStage_roomControl_c::getDemoArcName(), (u16)mModel.field_0x0.field_0x4);
+            if (anm == NULL) {
+                OS_REPORT("ESC_WARNING指定bckアニメーションが見つかりません！(%d)\n", mModel.field_0x0.field_0x4); // ESC_WARNING: Specified bck animation not found! (%d)
+                return 1;
+            }
+
+            fVar1 = 0.0f;
+            if (actor->checkEnable(0x80)) {
```
hatal175: Use ENABLE_ANM_TRANSITION_e

src/d/actor/d_a_demo00.cpp
```diff
@@ -4,1505 +4,1897 @@
...
+            mpBtkAnm->play();
+        } else if (mpBrkAnm != NULL) {
+            mpBrkAnm->play();
+        } else if (mpBpkAnm != NULL) {
+            mpBpkAnm->play();
+        } else if (mpBlkAnm != NULL) {
+            mpBlkAnm->play();
+        }
+
+        if (actor->checkEnable(4)) {
```
hatal175: Use ENABLE_SCALE_e

src/d/actor/d_a_demo00.cpp
```diff
@@ -4,1505 +4,1897 @@
...
     // NONMATCHING
+
+    field_0x6a1 = 0;
+    dDemo_actor_c* actor = dDemo_c::getActor(demoActorID);
+
+    if (actor == NULL) {
+        fopAcM_delete(this);
+        OS_REPORT("汎用くん<dactor%d>削除！！\n", subtype);
+    } else {
+        if (actor->checkEnable(0x10)) {
```
hatal175: use ENABLE_SHAPE_e

src/d/actor/d_a_demo00.cpp
```diff
@@ -4,1505 +4,1897 @@
...
+            mModel.field_0x0.mShapeID = actor->getShapeId();
+        }
+
+        if (actor->checkEnable(0x20)) {
```
hatal175: Use ENABLE_ANM_e

src/d/actor/d_a_demo00.cpp
```diff
@@ -4,1505 +4,1897 @@
...
+            mModel.field_0x0.field_0x4 = actor->getAnmId();
+        }
+
+        int piVar1, piVar2, piVar3;
+        u16 upVar1;
+        u8 upVar2;
+        if (actor->checkEnable(1)) {
```
hatal175: Please add an enum to Enable_e and use it

include/d/d_demo.h
carco-games: In actStandby, the assembly shows that the function is called without a class specifier, so I needed to declare the function in the header. However, after removing the inline keyword, I think that fixed it.

hatal175: It shouldn't have a static modifier as well - this is not a function with static storage.

hatal175 (approved):

PR #2541, opened 2025-07-18
d_a_kago almost matching

caseif:
`d_a_kago` is 99% matching with this PR. There's a single regalloc issue in `checkRoofHeight` - any input on fixing this would be much appreciated, although it seems quite stubborn based on my attempts so far.

### DIFF
--- a/include/d/actor/d_a_kago.h
+++ b/include/d/actor/d_a_kago.h
@@ -1,68 +1,72 @@
 #ifndef D_A_KAGO_H
 #define D_A_KAGO_H
 
-#include "f_op/f_op_actor_mng.h"
+#include "d/d_bg_s_acch.h"
+#include "d/d_cc_d.h"
+#include "d/d_cc_uty.h"
+#include "d/d_msg_flow.h"
 #include "d/d_path.h"
+#include "f_op/f_op_actor_mng.h"
 
 /**
  * @ingroup actors-unsorted
  * @class daKago_c
- * @brief Basket
+ * @brief Player-controlled Kargarok
  *
  * @details
  *
  */
 class daKago_c : public fopAc_ac_c {
 public:
-    /* 80849660 */ void getBckName(int);
+    /* 80849660 */ int getBckName(int);
     /* 808496F0 */ void setBck(int, u8, f32, f32);
-    /* 80849798 */ void checkBck(int);
-    /* 808497F8 */ void draw();
-    /* 80849980 */ void executeBalloonMenu();
-    /* 80849BA8 */ void checkGroundHeight(cXyz, f32*);
-    /* 8084A070 */ void checkRoofHeight(cXyz);
+    /* 80849798 */ bool checkBck(int);
+    /* 808497F8 */ int draw();
+    /* 80849980 */ int executeBalloonMenu();
+    /* 80849BA8 */ f32 checkGroundHeight(cXyz, f32*);
+    /* 8084A070 */ f32 checkRoofHeight(cXyz);
     /* 8084A210 */ void checkMoveHeight();
     /* 8084A2A8 */ void checkSizeBg();
-    /* 8084A6DC */ bool getBeforeGroundHeight(u8);
+    /* 8084A6DC */ s16 getBeforeGroundHeight(u8);
     /* 8084A6E4 */ void demo_skip(int);
-    /* 8084AA64 */ void DemoSkipCallBack(void*, int);
+    /* 8084AA64 */ static int DemoSkipCallBack(void*, int);
     /* 8084AA98 */ void setActionMode(int, int);
     /* 8084AAAC */ void setMidnaTagPos();
     /* 8084AB40 */ void setMidnaRideOn();
     /* 8084AB8C */ void setPlayerRideOn();
     /* 8084ABC8 */ void setRideOff();
-    /* 8084AC2C */ void searchNearPassPoint();
-    /* 8084AE78 */ void setSceneChange(int);
+    /* 8084AC2C */ s8 searchNearPassPoint();
+    /* 8084AE78 */ int setSceneChange(int);
     /* 8084AF40 */ void createBalloonScore();
     /* 8084AFB0 */ void startBalloonScore();
     /* 8084B024 */ void endBalloonScore();
     /* 8084B088 */ f32 checkNextPath(cXyz);
     /* 8084B108 */ void checkHeight();
-    /* 8084B134 */ void checkYaguraPos(cXyz);
-    /* 8084B1E0 */ void checkWallHitFall(int);
-    /* 8084B56C */ void checkAttackStart();
-    /* 8084B5DC */ void getValueY(f32);
-    /* 8084B6F4 */ void getValueX(f32);
-    /* 8084B718 */ void getValueAbsX(f32);
+    /* 8084B134 */ bool checkYaguraPos(cXyz);
+    /* 8084B1E0 */ bool checkWallHitFall(int);
+    /* 8084B56C */ bool checkAttackStart();
+    /* 8084B5DC */ s16 getValueY(f32);
+    /* 8084B6F4 */ s16 getValueX(f32);
+    /* 8084B718 */ s16 getValueAbsX(f32);
     /* 8084B744 */ void flySpeedCalcLockOn();
     /* 8084B96C */ void flySpeedCalc(s16, int);
-    /* 8084C078 */ void checkFlySceneChange();
+    /* 8084C078 */ bool checkFlySceneChange();
     /* 8084C178 */ void setFlyAway();
     /* 8084C204 */ void setFlyAnime();
     /* 8084C2CC */ void executeFly();
     /* 8084C8FC */ void executeStagger();
     /* 8084D108 */ void executeWait();
-    /* 8084D354 */ void calcAttackMove(int);
+    /* 8084D354 */ bool calcAttackMove(int);
     /* 8084D50C */ void executeAttack();
     /* 8084DA18 */ void calcCircleCamera(int);
     /* 8084DBFC */ void executeEvent();
     /* 8084DD18 */ void initPerchDemo();
-    /* 8084E2D4 */ void executePerchDemo();
-    /* 8084F09C */ void PerchDemoAwayForward();
+    /* 8084E2D4 */ bool executePerchDemo();
+    /* 8084F09C */ bool PerchDemoAwayForward();
     /* 8084FA58 */ void executePerch();
     /* 8084FAA8 */ void executeEvent2();
     /* 8084FB50 */ void initFirstDemo();
-    /* 80850384 */ void executeFirstDemo();
+    /* 80850384 */ bool executeFirstDemo();
     /* 808513E4 */ void executePerch2();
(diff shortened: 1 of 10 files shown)


include/d/actor/d_a_kago.h
```diff
@@ -1,68 +1,72 @@
...
+#include "d/d_cc_uty.h"
+#include "d/d_msg_flow.h"
 #include "d/d_path.h"
+#include "f_op/f_op_actor_mng.h"
 
 /**
  * @ingroup actors-unsorted
  * @class daKago_c
- * @brief Basket
+ * @brief Friendly Kargarok
```
TakaRikka: i'd add on to this something to explain it's the kargorok the player flies, like `player-controlled kargorok` or `flyable kargorok` or `controllable kargorok` etc.

include/d/actor/d_a_kago.h
```diff
@@ -80,67 +84,180 @@ class daKago_c : public fopAc_ac_c {
...
-    /* 0x6E1 */ u8 mIsFlying;
-    /* 0x6E2 */ u8 field_0x6E2[0x740 - 0x6E2];
-    /* 0x740 */ int field_0x740;
-    /* 0x744 */ int field_0x744;
-    /* 0x748 */ u8 field_0x748[0x76C - 0x748];
-    /* 0x76C */ dPath* field_0x76c;
-    /* 0x770 */ u8 field_0x770[0x773 - 0x770];
+    /* 0x650 */ cXyz field_0x650[3];
+    /* 0x674 */ cXyz field_0x674;
+    /* 0x680 */ cXyz field_0x680; ///
```
TakaRikka: can all these comments be removed?

include/d/actor/d_a_obj_riverrock.h
```diff
@@ -47,7 +47,9 @@ class daObjRIVERROCK_c : public dBgS_MoveBgActor {
 
 private:
     /* 0x5A0 */ ActionEnum mAction;
+public:
     /* 0x5A4 */ BreakSubAction mBreakSubAction;
+private:
```
TakaRikka: just make the first private into public and remove the other identifiers

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+        break;
+    }
+
+    dStage_changeScene(iVar2, 0.0f, 10, fopAcM_GetRoomNo(this), 0, -1);
 }
 
 /* 8084AF40-8084AFB0 001A80 0070+00 1/1 0/0 0/0 .text            createBalloonScore__8daKago_cFv */
 void daKago_c::createBalloonScore() {
-    // NONMATCHING
+    if (field_0x6e7 != 0 && mBalloon2DId == 0xffffffff) {
```
TakaRikka: mBalloon2DId should be of type `fpc_ProcID` and use the `fpcM_ERROR_PROCESS_ID_e` enum for the comparison

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+        mBalloon2DId = fopAcM_create(0x5a, 0, 0, 0xffffffff, 0, 0, 0xffffffff);
```
TakaRikka: `PROC_BALLOON2D` instead of 0x5a

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+        field_0x6e9 = 1;
+    }
 }
 
 /* 8084AFB0-8084B024 001AF0 0074+00 1/1 0/0 0/0 .text            startBalloonScore__8daKago_cFv */
 void daKago_c::startBalloonScore() {
-    // NONMATCHING
+    fopAc_ac_c* balloon_actor;
+    if ((field_0x6e7 != 0) && (field_0x6e9 == 1)) {
```
TakaRikka: remove inner parenthesis

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
 
 /* 8084B56C-8084B5DC 0020AC 0070+00 1/1 0/0 0/0 .text            checkAttackStart__8daKago_cFv */
-void daKago_c::checkAttackStart() {
-    // NONMATCHING
+bool daKago_c::checkAttackStart() {
+    if (mDashCooldownTime != 0) {
+        return false;
+    }
+
+    if (mDoCPd_c::getTrigA(0) != 0) {
```
TakaRikka: for all funcs that are returning either 0 or 1 i prefer to leave off the comparison to 0 if possible

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+
+                setActionMode(8, 0);
+                executeLanding();
+            }
+
+        }
+
+        return;
+
+        mPathDir = 1;
```
TakaRikka: whats going on here? this mPathDir set can't be reached with the return above

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+            setKagoPath(1);
+        } else {
+            setKagoPath(5);
+
+            if (field_0x73c == 1) {
+                createBalloonScore();
+            }
+        }
+
+        return;
```
TakaRikka: unnecessary explicit return

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
 void daKago_c::setWallHitEffect(cXyz param_0, int param_1) {
-    // NONMATCHING
+    static u16 kago_wall_hit_id[6] = {
+        0x8658,
+        0x86F9,
+        0x86FA,
+        0x8659,
+        0x86FB,
+        0x86FC,
+    };
```
TakaRikka, line 3427: this seems like a 2D array?

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+
+    if (field_0x6e7 == 1 && (mCurrentAction == 0 || mCurrentAction == 5)) {
+        mSph.OnAtSetBit();
+    }
+
+    setFlySound();
+
+    if (daPy_getPlayerActorClass()->checkCargoCarry()) {
+        dCamera_c* camera;
+        if ((mCurrentAction == 9) || (mCurrentAction == 10)) {
```
TakaRikka: remove inner parenthesis

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+            dCam_getBody()->SetTrimTypeForce(2);
+        } else {
+            if (isAttack()) {
+                dCam_getBody()->SetTrimTypeForce(0);
+                dCam_getBody()->ModeFix(5);
+            }
+        }
```
TakaRikka: this can probably be turned into an else if

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+        dComIfG_Ccsp()->Set(&mSph);
+    }
 }
 
 /* 80854058-80854168 00AB98 0110+00 1/1 0/0 0/0 .text            execute__8daKago_cFv */
-void daKago_c::execute() {
-    // NONMATCHING
+int daKago_c::execute() {
+    if (executeBalloonMenu()) {
+#if !DEBUG
```
TakaRikka: if this part isnt in debug, then i believe its more of a version difference rather than a debug thing. use the version check defines instead

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
-    // NONMATCHING
+/* 808544D8-80854928 00B018 0450+00 1/1 0/0 0/0 .text            create__8daKago_cFv */
+int daKago_c::create() {
+    fopAcM_SetupActor(this, daKago_c);
+
+    u32 param = (u8)fopAcM_GetParam(this);
+    if ((param & 0xff) != 0xff) {
+        mpPath1 = dPath_GetRoomPath(param & 0xff, fopAcM_GetRoomNo(this));
+        if (mpPath1 == NULL) {
+            OS_REPORT("");
```
TakaRikka: all of these OS_REPORTS seem like they're missing the correct string?

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+    }
+    if (field_0x6e7 == 0) {
+        field_0x75c = "E_YC";
+        field_0x760 = "KAGO_YAMI";
+    } else {
+        field_0x75c = "E_KC";
+        field_0x760 = "KAGO_HIKARI";
+    }
+
+    s32 loadResult = dComIfG_resLoad(&mPhase, field_0x75c);
```
TakaRikka: call this phase_state, and generally use int instead of s32

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+                    field_0x73c = 1;
+                }
+            } else {
+                if (strcmp("F_SP126", dComIfGp_getStartStageName()) == 0 &&
+                    dComIfGp_getStartStageRoomNo() == 0)
+                {
+                    field_0x73c = 2;
+                }
+            }
+        }
```
TakaRikka: these nests should be able to be turned into else ifs

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3967 @@
...
+
+        mBalloon2DId = -1;
```
TakaRikka: fpcM_ERROR_PROCESS_ID_e

src/d/actor/d_a_kago.cpp
caseif: Good observation, although I'm not able to come anywhere close to a match using a 2D array instead. Would it be ok to add a comment for now noting that it might be a fakematch?

hatal175 (approved):

src/d/actor/d_a_kago.cpp
```diff
@@ -4,1862 +4,3954 @@
...
+            unkInt2 = 0;
+        } else {
+            unkXyz1 = dPath_GetPnt(mpPath1, mPathIdx)->m_position;
+            unkXyz2 = dPath_GetPnt(mpPath1, mPathIdx - mPathIdxOffset)->m_position;
+            unkInt1 = cLib_targetAngleY(&unkXyz2, &unkXyz1);
+            unkInt2 = -cLib_targetAngleX(&unkXyz2, &unkXyz1);
+        }
+        if (field_0x744 == 3) {
+            cLib_chaseUC(&field_0x6de, 0, 4);
+            cLib_addCalcAngleS(&current.angle.x, 0xffffe000, 8, 0x100, 0x10);
```
YunataSavior, line 1207: A little late to be giving this comment, but here it goes:

Commenting here, but this appears throughout the TU: there are many spots where VSCode will flag these hex values as improper arguments. In reality, `0xffffe000` is `-0x2000`, and you should have put that there instead.

PR #2551, opened 2025-07-23
Matching d_a_obj_poFire

Stbn-antier:
100% match, weak function ordering issue.
95% match on debug as well.


src/d/actor/d_a_obj_poFire.cpp
```diff
@@ -1,602 +1,418 @@
...
 void daPoFire_c::modeFireEnd() {
-    // NONMATCHING
+    for (int i = 0; i < 2; i++) {
+        JPABaseEmitter* emitter = dComIfGp_particle_getEmitter(mParticleKeys[i]);
+        if (emitter != NULL) {
+            emitter->deleteAllParticle();
+            dComIfGp_particle_levelEmitterOnEventMove(mParticleKeys[i]);
+        }
+    }
+    fopAcM_delete((fopAc_ac_c*)this);
```
hatal175: conversion needed?

src/d/actor/d_a_obj_poFire.cpp
```diff
@@ -605,103 +421,85 @@ bool daPoFire_c::eventEnd() {
...
+// #endif
 
-/* 80CB3E58-80CB3ED8 0015F8 0080+00 2/1 0/0 0/0 .text            __dt__10daPoFire_cFv */
-daPoFire_c::~daPoFire_c() {
-    // NONMATCHING
-}
+/* 80CB4050-80CB4070 -00001 0020+00 1/0 0/0 0/0 .data            l_daPoFire_Method */
+static actor_method_class l_daPoFire_Method = {
+    (process_method_func)daPoFire_Create,  (process_method_func)daPoFire_Delete,
+    (process_method_func)daPoFire_Execute, 0,
```
hatal175: NULL instead of 0

src/d/actor/d_a_obj_poFire.cpp
```diff
@@ -605,103 +421,85 @@ bool daPoFire_c::eventEnd() {
...
+    fopAcM_RegisterCreateID(daPoFire_c, i_this, "daPoFire");
+    return a_this->create();
 }
 
 /* 80CB3E3C-80CB3E58 0015DC 001C+00 1/1 0/0 0/0 .text            getFirePos__12daPoCandle_cFv */
-void daPoCandle_c::getFirePos() {
-    // NONMATCHING
-}
+cXyz daPoCandle_c::getFirePos() {
+    return static_cast<cXyz>(mFirePos);
```
hatal175, line 463: Why do you need a static_cast here?

Stbn-antier: Debug asm had a call to `cXyz::cXyz(const cXyz&)`, returning a static cast is the only way I found to match debug

PR #2560, opened 2025-07-28
d_a_bd Matching

Stbn-antier:
100% text match, weak function ordering issue.
90% match on debug as well.


src/d/actor/d_a_bd.cpp
```diff
@@ -1,1225 +1,1117 @@
...
-        {0x0, {{0x0, 0x0, 0x0}, {0xd8fbfdff, 0x3}, 0x75}}, // mObj
-        {dCcD_SE_NONE, 0x0, 0x0, 0x0, 0x0}, // mGObjAt
-        {dCcD_SE_NONE, 0x0, 0x0, 0x0, 0x2}, // mGObjTg
-        {0x0}, // mGObjCo
-    }, // mObjInf
-    {
-        {{0.0f, 0.0f, 0.0f}, 10.0f} // mSph
-    } // mSphAttr
-};
+static u8 padding[0x3D];  // Padding to align .bss
```
hatal175: I think this should be UNK_REL_DATA since I don't see you used it but deleted all the lit_1109 block

src/d/actor/d_a_bd.cpp
```diff
@@ -1,1225 +1,1117 @@
...
+    f32 var_f31;
+
+    cXyz sp8 = i_this->field_0x5C4 - player->current.pos;
+    u32 var_r28 = 0;
+    if (i_this->field_0x658 != 0 || i_this->field_0x5B6 >= 1) {
+        return;
+    }
+
+    if ((i_this->field_0x618 & 0xF) == 0) {
+        var_r28 = (u32)fpcM_Search((fopAcIt_JudgeFunc)s_a_sub, a_this);
```
hatal175: Does this need the fopAcIt_JudgeFunc conversion?

src/d/actor/d_a_bd.cpp
```diff
@@ -1,1225 +1,1117 @@
...
+    }
+
+    if (daPy_getPlayerActorClass()->checkHorseRide()) {
+        var_f30 = dComIfGp_getHorseActor()->speedF;
+    } else {
+        var_f30 = player->speedF;
+    }
+
+    var_f31 = l_HIO.mLinkDetectRange + 10.0f * fabsf(var_f30);
+    if (player->getCutType() != 0) {
```
hatal175: How about we add CUT_TYPE_NONE to this enum since I see this comparison has been done in a few more places

src/d/actor/d_a_bd.cpp
```diff
@@ -1,1225 +1,1117 @@
...
+        }
+    }
+
+    a_this->speedF = speed;
+    if ((i_this->field_0x618 & 0xF) == 0 && cM_rndF(1.0f) < 0.35f) {
+        i_this->mSound.startSound(Z2SE_BIRD_STAY, 0, -1);
+    }
+    drop_check(i_this);
+    if (i_this->field_0x658 == 0) {
+        if (fopAcM_CheckCondition(a_this, fopAcCnd_NODRAW_e) != 0) {
```
hatal175: != 0 needed?

src/d/actor/d_a_bd.cpp
```diff
@@ -1,1225 +1,1117 @@
...
+    a_this->speedF = 0.0f;
+    a_this->gravity = 0.0f;
+    a_this->speed.y = 0.0f;
+    if (i_this->field_0x64C[1] == 0) {
+        if (cM_rndF(1.0f) < 0.2f) {
+            i_this->field_0x64C[1] = (20.0f + cM_rndF(10.0f));
+        } else {
+            i_this->field_0x64C[1] = (2.0f + cM_rndF(2.0f));
+        }
+        anm_init(i_this, wait_bck[(int)cM_rndF(1.99f)], 2.0f, 0, 1.0f);
```
hatal175, line 734: I don't think you need the (int) on all the cm_rndF calls that are used as indices. It's pretty much implicit.

src/d/actor/d_a_bd.cpp
```diff
@@ -1,1225 +1,1117 @@
...
-/* 804DA444-804DA448 00012C 0004+00 0/0 0/0 0/0 .bss sInstance__32JASGlobalInstance<11Z2FxLineMgr>
- */
-#pragma push
-#pragma force_active on
-static u8 data_804DA444[4];
-#pragma pop
-
-/* 804DA448-804DA44C 000130 0004+00 0/0 0/0 0/0 .bss sInstance__31JASGlobalInstance<10Z2EnvSeMgr>
- */
-#pragma push
```
hatal175, line 1206: I think you're missing AUDIO_INSTANCES

src/d/actor/d_a_bd.cpp
Stbn-antier: UNK_REL_BSS indeed improves the matching for .bss, thanks!

src/d/actor/d_a_bd.cpp
Stbn-antier: Unfortunately I get illegal operand errors without the explicit conversion

src/d/actor/d_a_bd.cpp
Stbn-antier: Sounds good, I added it in the enum and updated the comparison where it was done

hatal175 (approved):

PR #2572, opened 2025-08-06
Misc. framework/d/ improvements

caseif:
This PR fixes some mismatches under `framework/d` with ~~a couple of~~ one TU now fully matching. I noticed that other framework classes are marked as matching for JP, but I don't have a JP ROM at the moment so I can't test this for the new matches.


config/GZ2E01/symbols.txt
```diff
@@ -2031,7 +2031,7 @@ drawDoorCommon__19renderingPlusDoor_cFPC21stage_tgsc_data_classib = .text:0x8004
 checkDispDoorS__19renderingPlusDoor_cFiif = .text:0x80040710; // type:function size:0x128 scope:global align:4
 drawNormalDoorS__19renderingPlusDoor_cFPC21stage_tgsc_data_classiib = .text:0x80040838; // type:function size:0x17C scope:global align:4
 isDrawRoomIcon__16renderingDAmap_cCFii = .text:0x800409B4; // type:function size:0x2C scope:global align:4
-isDrawIconSingle__16renderingDAmap_cCFPCQ27dTres_c6data_siibbPC3Vec = .text:0x800409E0; // type:function size:0xB4 scope:global align:4
+isDrawIconSingle__16renderingDAmap_cCFPCQ27dTres_c6data_siiibPC3Vec = .text:0x800409E0; // type:function size:0xB4 scope:global align:4
```
TakaRikka: what is the reasoning for this change? these should've been taken straight from the original map file, which says its `isDrawIconSingle__16renderingDAmap_cCFPCQ27dTres_c6data_siibbPC3Vec`

include/d/d_demo.h
```diff
@@ -164,10 +164,10 @@ class dDemo_system_c : public JStage::TSystem {
 public:
     dDemo_system_c() { mpObject = NULL; }
 
-    /* 80039AAC */ virtual ~dDemo_system_c();
     /* 80039528 */ virtual int JSGFindObject(JStage::TObject**, char const*,
                                               JStage::TEObject) const;
-    
+    /* 80039AAC */ virtual ~dDemo_system_c();
```
TakaRikka, line 169: are you sure all these virtuals should move? make sure it doesn't affect the vtable's function order

src/d/d_demo.cpp
```diff
@@ -255,7 +255,9 @@ int dDemo_actor_c::getDemoIDData(int* param_0, int* param_1, int* param_2, u16*
     return 1;
 }
 
-s16 dDemo_c::m_branchNum = 0;
+//#if DEBUG
+//s16 dDemo_c::m_branchNum = 0;
+//#endif
```
TakaRikka: why commented out?

config/GZ2E01/symbols.txt
caseif: This was previously done in #2201 before being reverted in #2203 a few days later. I assumed the reversion was just an oversight, but if the map file is a source of truth then I'll undo this since it obviously wouldn't be the correct way to match.

include/d/d_demo.h
caseif: This was the only way that I could find to fix virtual function ordering in the TU. As far as I can tell (based on the `.data` section of this TU as well as virtual function calls in `.text`) it shouldn't have broken vtable ordering, but I'm not sure if there's anything else I should check to verify.

src/d/d_demo.cpp
caseif: My bad, I meant to uncomment this before committing.

config/GZ2E01/symbols.txt
TakaRikka: yeah map file should be followed as much as possible preferably. could try making it `const bool` to see if it helps, since adding const to primitive types doesn't actually change the function signature but can effect things like regalloc

caseif: Does the map file come from the debug ROM? I'm wondering if there's a version difference that could explain it since changing it to `BOOL` fixes the mismatches so nicely.

TakaRikka: the maps are in all versions of the game except korean wii and HD, in the maps folder. there is no version difference for that symbol as far as i can tell

caseif: Ah okay, in that case I'm pretty stumped so I'll just leave it as I found it.

TakaRikka (approved):

PR #2578, opened 2025-08-11
Misc framework fixes

caseif:
This PR contains some more miscellaneous work on framework TUs, including a pretty iffy hack to get `m_Do_graphic` matching. There are also a couple of changes to fix debug compilation.

### DIFF
--- a/include/d/actor/d_a_npc_kolin.h
+++ b/include/d/actor/d_a_npc_kolin.h
@@ -67,6 +67,9 @@ public:
     /* 80558AF8 */ int follow(void*);
     /* 80558C4C */ int clothWait(void*);
     /* 80558E90 */ int talk(void*);
+#if DEBUG
+    int test(void*);
+#endif
     /* 8055A3B8 */ daNpc_Kolin_c(
             daNpcT_faceMotionAnmData_c const* i_faceMotionAnmData,
             daNpcT_motionAnmData_c const* i_motionAnmData,
--- a/include/d/actor/d_a_npc_kolinb.h
+++ b/include/d/actor/d_a_npc_kolinb.h
@@ -115,6 +115,9 @@ public:
     /* 80A4738C */ int cutThankYou(int);
     /* 80A4742C */ int wait(void*);
     /* 80A475D4 */ int talk(void*);
+#if DEBUG
+    int test(void*);
+#endif
     /* 80A4854C */ daNpc_Kolinb_c(
             daNpcT_faceMotionAnmData_c const* i_faceMotionAnmData,
             daNpcT_motionAnmData_c const* i_motionAnmData,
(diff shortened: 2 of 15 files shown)


include/d/actor/d_a_npc_kolin.h
```diff
@@ -67,6 +67,7 @@ class daNpc_Kolin_c : public daNpcT_c {
     /* 80558AF8 */ int follow(void*);
     /* 80558C4C */ int clothWait(void*);
     /* 80558E90 */ int talk(void*);
+    int test(void*);
```
TakaRikka, line 71: is this real?

include/d/actor/d_a_npc_kolinb.h
```diff
@@ -115,6 +115,7 @@ class daNpc_Kolinb_c : public daNpcT_c {
     /* 80A4738C */ int cutThankYou(int);
     /* 80A4742C */ int wait(void*);
     /* 80A475D4 */ int talk(void*);
+    int test(void*);
```
TakaRikka, line 119: is this real?

include/d/d_cc_d.h
```diff
@@ -420,6 +423,34 @@ class dCcD_GObjInf : public cCcD_GObjInf {
     /* 0x0E8 */ dCcD_GObjCo mGObjCo;
 };  // Size = 0x104
 
+// Point (unused, existence is inferred)
```
TakaRikka, line 426: is this for virtual ordering stuff? since it's existence is inferred, i'd add a note about it's purpose just incase it turns out that its not real

include/m_Do/m_Do_graphic.h
```diff
@@ -4,6 +4,9 @@
 #include "JSystem/JFramework/JFWDisplay.h"
 #include "dolphin/mtx.h"
 #include "global.h"
+#pragma sym off
```
TakaRikka, line 10: i'd add a note about this hack since yeah its kinda sketch

src/m_Do/m_Do_graphic.cpp
```diff
@@ -17,7 +17,9 @@
 #include "m_Do/m_Do_machine.h"
 #include "m_Do/m_Do_main.h"
 #include "SSystem/SComponent/c_math.h"
+#pragma sym off
```
TakaRikka, line 23: same thing about making a note here

include/d/actor/d_a_npc_kolin.h
caseif: Yep, it's present in debug only.

include/d/d_cc_d.h
caseif: After testing again this doesn't actually seem to have an effect on function ordering per se, but it does cause the `cCcD_PntAttr` definitions to be emitted in `d_cc_d` in debug and seems to be the most plausible way to achieve this. That makes sense though, I can add a comment noting this.

include/d/actor/d_a_npc_kolin.h
YunataSavior: You may want to put `#if DEBUG` macros around this just to make it clear.

TakaRikka (approved):

PR #2584, opened 2025-08-13
d_a_obj_key / d_a_obj_keyhole / d_a_obj_smallkey

TakaRikka:


src/d/actor/d_a_obj_smallkey.cpp
```diff
@@ -189,670 +26,590 @@ const static dCcD_SrcCyl l_cyl_src = {
...
-    // NONMATCHING
-}
+static f32 Reflect(cXyz* i_vec, cBgS_PolyInfo const& i_polyinfo, f32 i_scale) {
+    cM3dGPla plane;
+
+    if (dComIfG_Bgsp().GetTriPla(i_polyinfo, &plane)) {
+        cXyz reflect_vec;
+        f32 mag = i_vec->absXZ();
+
+        C_VECReflect(i_vec, &plane.mNormal, &reflect_vec);
```
hatal175: Use VECReflect macro (I know it's the same, but well, either we remove the macro or use it consistently)

PR #2597, opened 2025-08-22
Add precompiled headers for framework and RELs

caseif:
This PR adds a precompiled header (`dolzel.pch`) for use in RELs to match the original build process. This has a huge effect on weak function ordering and as a result many TUs are now matching as a result of these changes. This also seems to solve (at least partly) the `Z2Calc::cNullVec`, `lit_1787`, and `lbl_*_data_10` `.data` entries and `lit_1109`-`lit_1009` `.bss` entries seen across the codebase.

This PR also removes `nosyminline on` from the default REL flags (and from many individual framework objects) since it is no longer necessary in the majority of cases.

One issue with the changeset in its current state involves the handling of the `sqrtf` constants. The PCH needs to be built with `.sdata` and `.sdata2` enabled in order for `d_kankyo` and `d_resorce` to match; however, this causes the constants to be placed in `.sdata2` even in RELs such as in `d_a_b_yo` or `d_a_L7op_demo_dr` (where they should be in `.rodata`). This not only breaks matches, but also breaks linking entirely with this message:

```
#   Small data sections must have their own output sections specified in the 
#   linker command file.
#   No linker command file output section '.sdata2'.
```

The solution to this for now is to build the PCH twice: once with framework flags, and once with REL flags. Each object links one version or the other as appropriate. This may or may not be a fakematch and should be replaced if a better solution is found in the future.

This PR includes the changes from https://github.com/encounter/dtk-template/pull/60 in order to support building the PCHs. objdiff isn't yet capable of understanding the weak data stripping going on here, so there are a LOT of false data regressions reported as a result.

See https://github.com/zeldaret/tww/pull/860 for the corresponding Wind Waker PR for PCH changes.


include/weak_data_lit_1787.h
```diff
@@ -0,0 +1,8 @@
+#ifndef WEAK_DATA_LIT_1787_H
+#define WEAK_DATA_LIT_1787_H
+
+#include "dolphin/types.h"
+
+static u32 lit_1787 = 0x02000201;
```
TakaRikka: wasnt this supposed to be some J3DJoint static data or something like that? lago said it was something like that for tww

src/d/actor/d_a_bg_obj.cpp
```diff
@@ -256,6 +72,10 @@ static const char* getDzbName(int param_0) {
     return l_dzbName;
 }
 
+static const char* dummy() {
```
TakaRikka, line 75: is the dummy necessary?

src/d/d_stage.cpp
```diff
@@ -3,6 +3,10 @@
 // Translation Unit: d/d_stage
 //
 
+//#include "d/dolzel_framework.h"
```
TakaRikka: unnecessary comment?

src/d/d_stage.cpp
```diff
@@ -1601,13 +1617,21 @@ static int dStage_cameraCreate(stage_camera2_data_class* i_cameraData, int i_cam
...
+}
+
+/*static void dummy() {
     // Needed to fix weak function order.
     // This is likely caused by the dStage_chkPlayerId function from TWW using these functions.
     // dStage_chkPlayerId isn't used in TP, so it gets stripped out, but the effect it has on weak order remains.
     dComIfGp_getStage()->getPlayer();
     dComIfGp_roomControl_getStatusRoomDt(0)->getPlayer();
-}
+}*/
```
TakaRikka: is this section able to be removed if it's commented out?

configure.py
```diff
@@ -533,44 +553,44 @@ def MatchingFor(*versions) -> bool:
         "host": True,
         "objects": [
             Object(NonMatching, "d/d_home_button.cpp"),
-            Object(MatchingFor("GZ2E01", "GZ2J01"), "d/d_stage.cpp", extra_cflags=['-pragma "nosyminline on"']),
-            Object(MatchingFor("GZ2E01", "GZ2J01"), "d/d_map.cpp"),
-            Object(MatchingFor("GZ2E01", "GZ2J01"), "d/d_com_inf_game.cpp", extra_cflags=['-pragma "nosyminline on"']),
+            Object(MatchingFor("GZ2E01", "GZ2J01"), "d/d_stage.cpp"),
+            Object(Equivalent, "d/d_map.cpp"), # weak func order
```
TakaRikka, line 557: are these ones that were changed from matching to equivalent not compatible with this?

src/d/actor/d_a_bg_obj.cpp
caseif: It's there to force it to be placed earlier in `.rodata`, not sure if there's a more elegant way to do this though.

include/weak_data_lit_1787.h
LagoLunatic: Yeah it's `getAttnFn` which is supposed to go in `J3DMatBlock.h`

src/d/d_stage.cpp
LagoLunatic: This dummy function I put here was closer to the realmatch than putting `getPlayer` directly in the cpp like is done in this PR. They're weak functions that are intended to go in the header, and then this unused function leftover from tww puts them in this location. You can fakematch them by making them non-weak, but it would be more accurate to put them in the header.

configure.py
caseif: For `d_map` including the PCH completely messes up the function order - I think that this object may actually not have included it. For `d_select_cursor` I think the PCH is correct, but it messes up the vtable padding which was previously solved with a fakematch (which I think ultimately comes down to a vtable ordering issue). I'll revert the first change for now and add a comment that it might be fake - let me know if you want me to revert the second change as well.

src/d/d_stage.cpp
caseif: Does the same go for the other functions moved from the header in this PR?

LagoLunatic: Yes, the virtual get/set functions in d_stage.h are all weak.

For reference in tww I was able to match all of their order in d_stage.h properly with the pch and no fakematches, though it was more difficult than most TUs due to some strange behavior with them being inherited from an abstract base class (`dStage_dt_c`).

include/JSystem/JAudio2/JASSimpleWaveBank.h
```diff
@@ -7,7 +7,7 @@
 
 struct JASSimpleWaveBank : JASWaveBank, JASWaveArc {
     struct TWaveHandle : JASWaveHandle {
-        /* 80298D84 */ ~TWaveHandle() {}
+        /* 80298D84 */ ~TWaveHandle();
```
LagoLunatic: You can just delete this line entirely instead of adding it to the .cpp. It's a weak function but defining ctors/dtors explicitly in the header results in them getting their own .text section split, but not defining them at all and letting the compiler add an implicit one merges it up into an earlier section.

src/d/actor/d_a_b_mgn.cpp
```diff
@@ -1,18 +1,34 @@
 /**
- * @file d_a_b_mgn.cpp
+* @file d_a_b_mgn.cpp
```
LagoLunatic: Accidentally got rid of the indent

configure.py
```diff
@@ -419,6 +427,18 @@ def MatchingFor(*versions) -> bool:
 
 config.warn_missing_config = True
 config.warn_missing_source = False
+config.precompiled_headers = [
+    {
+        "source": "d/dolzel_framework.pch",
```
LagoLunatic: I would name the framework one like this so the vibration symbols match in the debug build.
```suggestion
        "source": "d/dolzel.pch",
```

TakaRikka (approved):

PR #2602, opened 2025-08-24
d_a_obj_mie OK

hatal175:


YunataSavior:
@hatal175 , if you tried marking the TU as Matching, would it go through?

hatal175:
I've no idea what segfaulted in the github build

TakaRikka:
is the pch data actually being linked? cause objdiff is just going to show it for any TU that includes it now, making it a bit annoying to diff, but it shouldnt actually effect anything when trying to build due to it being stripped

YunataSavior:
@hatal175 , in `daObj_Mie_c::setWaterPrtcl`, change `emttrId[8]` to `emttrId[4]`. It will then work.

PR #2615, opened 2025-08-29
d_a_obj_kbox matching

hatal175:
Also small change to dKy_darkworld_check


hatal175:
Strange... was sure that i checked that. I will provably revert the dark world change later.

PR #2617, opened 2025-08-29
d_a_e_ai done

TakaRikka:


src/d/actor/d_a_e_ai.cpp
```diff
@@ -6,792 +6,994 @@
...
-    // NONMATCHING
+BOOL e_ai_class::player_way_check() {
+    s16 angle_diff = shape_angle.y - dComIfGp_getPlayer(0)->shape_angle.y;
+    if (angle_diff < 0) {
+        angle_diff = -angle_diff;
+    }
+
+    if ((u16)angle_diff < 0x4000) {
+        return FALSE;
+    } else {
```
hatal175: Can remove else (What is our policy on this?)

src/d/actor/d_a_e_ai.cpp
```diff
@@ -6,792 +6,994 @@
...
+    end = i_other->current.pos;
+    end.y += 100.0f;
+
+    start = current.pos;
+    start.y = eyePos.y;
+
+    linchk.Set(&start, &end, this);
+    if (dComIfG_Bgsp().LineCross(&linchk)) {
+        return TRUE;
+    } else {
```
hatal175: same

PR #2618, opened 2025-08-30
Partial Work on e_ww (55% debug match)

Stbn-antier:
Partial decomp of e_ww, still work in progress but I don't have time to complete it.
Doesn't include recent changes like .pch and not cleaned up.


TakaRikka:
wip is fine but the conflicts need to be resolved before merging

hatal175 (approved):

PR #2625, opened 2025-09-03
Link a lot of TUs for JP version

caseif:
This PR links a bunch of TUs for `GZ2J01`. This is almost entirely low-hanging fruit (i.e. TUs that were already matching and just needed to be marked or only required minor tweaks).


YunataSavior:
Cool stuff! As we begin linking other versions, it'll be interesting to see what all changes the Nintendo devs did between builds.

hatal175 (approved):

PR #2664, opened 2025-09-16
d_a_npc_grz matching

carter-ktb21:


src/d/actor/d_a_npc_grz.cpp
```diff
@@ -6,2594 +6,2600 @@
...
+    mCyl1.Set(mCcDCyl);
+    mCyl1.SetStts(&mCcStts);
+    mCyl1.SetTgHitCallback(tgHitCallBack);
+    mCyl2.Set(mCcDCyl);
+    mCyl2.SetStts(&mCcStts);
+
+    for (int i = 0; i < 4; i++) {
+        mSphs[i].Set(mCcDSph);
+        mSphs[i].SetStts(&mCcStts);
+        mSphs[i].SetAtSe(0xB);
```
hatal175: dCcD_SE_HARD_BODY

include/d/d_event_manager.h
```diff
@@ -31,12 +31,12 @@ class dEvent_exception_c {
 #define EVENT_LIST_NUM 11
 #endif
 
+enum Event_Cut_Nums {
+    /* 0x7 */ NUM_EVT_CUTS_e = 0x7,
+};
```
TakaRikka: i dont think this change is correct, or rather half-finished. i think this should be local to the files it's being used in and the value changes per file. like in npc_grr, it should be set to 1 like it was before

src/d/actor/d_a_npc_grz.cpp
```diff
@@ -6,2594 +6,2594 @@
...
+        return true;
+    }
+
+    if (setBckAnm(anm, 1.0f, i_attr, 0, -1, i_modify)) {
+        mAnmFlags |= ANM_PLAY_BCK | ANM_PAUSE_BCK;
+        mExpressionLoops = 0;
+        return true;
+    }
+
+    OS_REPORT("%s: 表情Bckアニメーションの登録に失敗しました！\n", "d_a_npc_grz.cpp");
```
TakaRikka: you can use the `__FILE__` macro instead of manually specifying a filename

src/d/actor/d_a_npc_grz.cpp
```diff
@@ -6,2594 +6,2594 @@
...
+            }
+        } else if (msgTimer != 0 && !field_0x9eb) {
+            setExpressionTalkAfter();
+        }
+
+        switch (mExpression) {
+            case EXPR_LAUGH:
+            case EXPR_SNIFF:
+                if (mExpressionPhase != 0) {
+                    mAnmFlags &= 0xFFFFEFFF;
```
TakaRikka: ~0x1000, or whatever anm flag enum is 0x1000

PR #2667, opened 2025-09-18
d_a_e_wb equivalent

hatal175:


Pheenoh (approved):
good work on the demo_camera 🫡

PR #2668, opened 2025-09-18
d_a_e_wb actually matches

hatal175:


hev7:
This matches for GZ2P01 as well.

PR #2671, opened 2025-09-19
Link all oustanding US-matched TUs for JP

caseif:
This PR matches all TUs for GZ2J01 which were already matched for GZ2E01, and also includes some fixes that improve both versions.


include/d/d_save.h
```diff
@@ -504,6 +504,8 @@ class dSv_player_config_c {
     void setVibration(u8 i_status);
     u8 getPalLanguage() const;
 
+    u8 getUnk0() { return unk0; }
+    void setUnk0(u8 i_unk0) { unk0 = i_unk0; }
```
TakaRikka, line 509: add a note here (and the dComIfGs_ inline) that the real names are unknown so we can try to make a better name later on

src/Z2AudioLib/Z2SoundObjMgr.cpp
```diff
@@ -187,15 +187,25 @@ void Z2SoundObjMgr::searchEnemy() {
...
+#endif
+        case Z2_ENEMY_DN: // 0x0C
+        case Z2_ENEMY_DB: // 0x16
+        case Z2_ENEMY_GE: // 0x1F
+        case Z2_ENEMY_OC: // 0x22
+        case Z2_ENEMY_MM: // 0x26
+        case Z2_ENEMY_RB: // 0x29
+        case Z2_ENEMY_HB: // 0x2B
+        case Z2_ENEMY_TT: // 0x30
+        case Z2_ENEMY_SM2: // 0x35
```
TakaRikka: are these comments necessary?

src/d/d_meter_button.cpp
```diff
@@ -1474,6 +1474,33 @@ void dMeterButton_c::screenInitText() {
     mpTmRoot_c = new CPaneMgr(mpTextScreen, 'mg_null', 0, NULL);
     JUT_ASSERT(0, mpTmRoot_c != 0);
 
+#if VERSION == VERSION_GCN_JPN
+    if (g_dComIfG_gameInfo.info.getPlayer().getConfig().getUnk0() == 0) {
```
TakaRikka: this should probably be a dComIfGs_ inline

caseif:
Noted issues should be fixed!

PR #2692, opened 2025-09-24
d_a_obj_smgdoor OK

Pheenoh:
PAL/JPN don't match for some reason


src/d/actor/d_a_obj_smgdoor.cpp
```diff
@@ -1,483 +1,486 @@
...
+    cXyz cStack_84(player->attention_info.position);
+    cStack_84.y = player->current.pos.y;
+    acStack_78 = cStack_84 - current.pos;
+    f32 dVar4 = acStack_78.abs2XZ();
+    cXyz cStack_90;
+    cStack_90.set(cM_ssin(home.angle.y), 0.0f, cM_scos(home.angle.y));
+
+    if (dVar4 > 62500.0f) {
+        return 0;
+    } else {
```
hatal175: Could you reduce the else stack here since it's all returns?

src/d/actor/d_a_obj_smgdoor.cpp
```diff
@@ -1,483 +1,486 @@
...
-    // NONMATCHING
+int daObjSmgDoor_c::checkOpen() {
+    if (getSwbit() != 0xff) {
+        if (fopAcM_isSwitch(this, getSwbit()) == 0) {
+            return 0;
+        }
+    }
+
+    if (daPy_py_c::checkNowWolf()) {
+        return 0;
```
hatal175, line 398: remove else

PR #2703, opened 2025-09-27
Misc matches and fixes

LagoLunatic:


src/dolphin/gx/GXTexture.c
```diff
@@ -1,5 +1,6 @@
 #include <dolphin/gx.h>
 #include <dolphin/os.h>
+#include <string.h>
```
TakaRikka, line 3: whats the reasoning for the changes to dolphin stuff? just wanna know incase it needs to be upstreamed to the dolsdk repo

LagoLunatic: Just cleaning up a bunch of clangd errors such as:
`Call to undeclared function 'memset'; ISO C99 and later do not support implicit function declarations`
memset is declared in string.h, while the intrinsics like __cntlzw and __rlwimi are declared in global.h.

PR #2708, opened 2025-09-28
d_a_obj_ss_drink equivalent

Pheenoh:


src/PowerPC_EABI_Support/MSL/MSL_C/MSL_Common/Include/float.h
```diff
@@ -73,6 +73,8 @@ inline int __fpclassifyd(double __value) {
 #define FLT_MAX_EXP    128
 #define FLT_MAX_10_EXP 38
 
+#define FLT_MIN -1e+09
```
TakaRikka: this should probably be called something like dBgS_NO_GROUND or dBgS_INVALID_FLOOR and probably be put into d_bg_s.h or d_bg_s_gnd_chk.h

src/d/actor/d_a_obj_ss_drink.cpp
```diff
@@ -1,699 +1,593 @@
...
+}
 
-/* 80CE6B40-80CE6B4C 00017C 000C+00 2/2 0/0 0/0 .data            __vt__10dCcD_GStts */
-SECTION_DATA extern void* __vt__10dCcD_GStts[3] = {
-    (void*)NULL /* RTTI */,
-    (void*)NULL,
-    (void*)__dt__10dCcD_GSttsFv,
-};
+/* 80CE53EC-80CE55A8 0004EC 01BC+00 1/1 0/0 0/0 .text            CreateHeap__15daObj_SSDrink_cFv */
+// NONMATCHING - regalloc
```
TakaRikka: comment can be removed if its matching now

src/d/actor/d_a_obj_ss_drink.cpp
```diff
@@ -1,699 +1,593 @@
...
+}
 
-/* 80CE6B4C-80CE6B58 000188 000C+00 2/2 0/0 0/0 .data            __vt__12dBgS_AcchCir */
-SECTION_DATA extern void* __vt__12dBgS_AcchCir[3] = {
-    (void*)NULL /* RTTI */,
-    (void*)NULL,
-    (void*)__dt__12dBgS_AcchCirFv,
-};
+/* 80CE55F0-80CE5624 0006F0 0034+00 1/1 0/0 0/0 .text            Delete__15daObj_SSDrink_cFv */
+// idk
```
TakaRikka: can this comment be removed?

src/d/actor/d_a_obj_ss_drink.cpp
```diff
@@ -1,699 +1,593 @@
...
-/* 80CE6B64-80CE6B70 0001A0 000C+00 3/3 0/0 0/0 .data            __vt__8cM3dGCyl */
-SECTION_DATA extern void* __vt__8cM3dGCyl[3] = {
-    (void*)NULL /* RTTI */,
-    (void*)NULL,
-    (void*)__dt__8cM3dGCylFv,
-};
+    int chk = chkEvent();
+    if (chk) {
+        if (field_0xaf0) {
+            (this->*field_0xaf0)((void*)0);
```
TakaRikka: should be able to just pass in NULL

src/d/actor/d_a_obj_ss_drink.cpp
```diff
@@ -1,699 +1,593 @@
...
+    if (sVar2 != 0xffff && sVar2 != 0) {
+        bVar1 = false;
+    }
+#else
+    if (sVar2 == 0xffff || sVar2 == 0) {
+        bVar1 = true;
+    }
+#endif
+
+    return bVar1 ? -1 : sVar2;
```
TakaRikka: this kinda seems like
```c++
u16 var_r31 = home.angle.x;
return var_r31 == 0xFFFF || var_r31 == 0 ? -1 : var_r31;
```

src/d/actor/d_a_obj_ss_drink.cpp
```diff
@@ -1,699 +1,593 @@
...
+    J3DModelData* modelData = mpModel->getModelData();
+    fopAcM_SetMtx(this, mpModel->getBaseTRMtx());
+    fopAcM_setCullSizeBox(this, -30.0f, -15.0f, -30.0f, 30.0f, 45.0f, 30.0f);
+    eventInfo.setArchiveName(getResName());
+    attention_info.flags = 0;
+#ifdef DEBUG
+    attention_info.distances[7] = 6;
+#else
+    attention_info.distances[4] = 6;
+#endif
```
TakaRikka: using the fopAc_attn_X_e enums here will let you remove the ifdef block

src/d/actor/d_a_obj_ss_drink.cpp
```diff
@@ -1,699 +1,593 @@
...
+
+    if (field_0xaf0) {
+        ret = (this->*field_0xaf0)((void*)0);
+    }
+
+    field_0xb08 = 0;
+    field_0xaf0 = param_0;
+
+    if (field_0xaf0) {
+        ret = (this->*field_0xaf0)((void*)0);
```
TakaRikka: same thing about the void 0's here

src/d/actor/d_a_obj_ss_drink.cpp
```diff
@@ -1,699 +1,593 @@
...
+    }
+
+    gravity = -5.0f;
+
+#ifdef DEBUG
+    bool bVar1 = false;
+    bool bVar2 = false;
+#endif
+
+    if (daPy_getPlayerActorClass()->getGrabActorID() != -1) {
```
TakaRikka: since its an actor id use fpcM_ERROR_PROCESS_ID_e instead of -1

src/d/actor/d_a_obj_ss_drink.cpp
```diff
@@ -1,699 +1,593 @@
...
+        if (getParentPtr() != NULL) {
+            fopAc_ac_c* parent = (fopAc_ac_c*)getParentPtr();
+            attention_info.position.set(parent->attention_info.position);
+        }
+
+        ret = 0;
+
+        if (eventInfo.checkCommandTalk()) {
+            if (checkProcess(&daObj_SSDrink_c::talk)) {
+                ret = (this->*field_0xaf0)((void*)0);
```
TakaRikka: same thing about void 0

src/d/actor/d_a_obj_ss_drink.cpp
```diff
@@ -1,699 +1,593 @@
...
+    if (attention_info.flags == 0x40000040) {
+#else
+    if (attention_info.flags == 0x40000008) {
+#endif
+        attention_info.distances[1] = 0x6;
+#ifdef DEBUG
+        attention_info.distances[6] = 0x6;
+#else
+        attention_info.distances[3] = 0x6;
+#endif
```
TakaRikka: use fopAc_attn enums for the distance indices and fopAc_attn_flag enums for the flags

src/d/actor/d_a_obj_ss_drink.cpp
```diff
@@ -1,699 +1,593 @@
...
+int daObj_SSDrink_c::wait(void* param_0) {
+#ifdef DEBUG
+    switch (field_0xb08) {
+    case 0:
+    case 2:
+    case 1:
+        break;
+    }
+#endif
+    return 1;
```
TakaRikka, line 459: im betting if you put the return 1 into the switch statement, it'll match debug but get optimized into return 1 on retail, cleanly matching both

src/PowerPC_EABI_Support/MSL/MSL_C/MSL_Common/Include/float.h
YunataSavior: I'm unsure whether I agree or not. We have debug asserts where the value is explicitly typed out, so I question if the original devs did that or not.

TakaRikka: changing the JUT_ASSERT macro to a double define like tww did lets you put things like NULL or this float macro in while it expands the value. just forgot to ever update the macro, but it's probably the correct way to handle it

LagoLunatic: The official name is `G_CM3D_F_INF`. We know this from the demo version of TWW as it's a constant that appears in the symbol map for `c_m3d.cpp` in that version.

In all other versions it's a define instead of a constant, so in asserts it simply shows as `(1000000000.0f)` because JUT_ASSERT is a nested macro that converts defines into their value. Although for TP I haven't gotten around to fixing JUT_ASSERT to do that properly yet, so it won't work in asserts at the moment.

Also, the negative version isn't a separate define from the positive one, the original devs typed `-G_CM3D_F_INF` instead of `FLT_MIN` when checking for invalid ground.

src/d/actor/d_a_obj_ss_drink.cpp
Pheenoh: no, it generates extra instructions on retail

include/SSystem/SComponent/c_bg_s_gnd_chk.h
```diff
@@ -5,6 +5,8 @@
 #include "SSystem/SComponent/c_bg_s_poly_info.h"
 #include "SSystem/SComponent/c_xyz.h"
 
+#define dBgS_GROUND_MIN_H -1e+09f
```
TakaRikka: sorry since lago corrected me on the correct define for this, it should be
```
#define G_CM3D_F_INF (1000000000.0f)
```
in c_m3d.h

PR #2715, opened 2025-09-30
d_a_obj_pumpkin OK

### LINKED ISSUE #831: d_a_obj_pumpkin.cpp

YunataSavior:
In addition, please review what I've done for `mpHIO` in npc_jagar and obj_pumpkin. This is how we must work with l_HIO in these files moving forward, as that means of writing is how we can cleanly match both release and debug.

Closes #831

### DIFF
--- a/include/d/actor/d_a_npc_jagar.h
+++ b/include/d/actor/d_a_npc_jagar.h
@@ -14,7 +14,7 @@
 
 struct daNpc_Jagar_HIOParam {
     /* 0x00 */ daNpcT_HIOParam common;
-    /* 0x8C */ f32 pumpkin_monitoring_dist;
+    /* 0x8C */ f32 pumpkin_watch_range;
     /* 0x90 */ f32 pumpkin_watch_Ymax;
     /* 0x94 */ f32 pumpkin_watch_Ymin;
     /* 0x98 */ f32 running_speed;
@@ -38,6 +38,10 @@ public:
 
     daNpc_Jagar_HIOParam m;
 };
+
+#define NPC_JAGAR_HIO_CLASS daNpc_Jagar_HIO_c
+#else
+#define NPC_JAGAR_HIO_CLASS daNpc_Jagar_Param_c
 #endif
 
 class daNpc_Jagar_c : public daNpcT_c {
@@ -205,7 +209,7 @@ public:
         if (mType == 0 && fopAcM_GetName(pumpkin_p) == PROC_OBJ_PUMPKIN && !daNpcT_chkEvtBit(0xAE)
             && daNpcT_chkEvtBit(0x25B) && !daNpcT_chkEvtBit(0x235)) {
             ++field_0xffc;
-            reg_r30 = chkPointInArea(pumpkin_p->current.pos, current.pos, mpHIO->m.pumpkin_monitoring_dist,
+            reg_r30 = chkPointInArea(pumpkin_p->current.pos, current.pos, mpHIO->m.pumpkin_watch_range,
                                      mpHIO->m.pumpkin_watch_Ymax, mpHIO->m.pumpkin_watch_Ymin, 0);
             if (reg_r30) {
                 if (field_0xffc >= 4) {
@@ -222,11 +226,7 @@ public:
     static char* mCutNameList[7];
     static cutFunc mCutList[7];
 private:
-#if DEBUG
-    /* 0x0E40 */ daNpc_Jagar_HIO_c* mpHIO;
-#else
-    /* 0x0E40 */ daNpc_Jagar_Param_c* mpHIO;
-#endif
+    /* 0x0E40 */ NPC_JAGAR_HIO_CLASS* mpHIO;
     /* 0x0E44 */ dCcD_Cyl mCyl1;
     /* 0x0F80 */ u8 mType;
     /* 0x0F84 */ daNpcT_ActorMngr_c mActorMngr[5];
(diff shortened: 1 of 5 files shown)


include/d/actor/d_a_npc_jagar.h
```diff
@@ -12,58 +12,33 @@
  *
  */
 
+struct daNpc_Jagar_HIOParam {
+    /* 0x00 */ daNpcT_HIOParam common;
+    /* 0x8C */ f32 pumpkin_monitoring_dist;
```
TakaRikka: `pumpkin_watch_range`

include/d/actor/d_a_npc_jagar.h
```diff
@@ -225,10 +200,33 @@ class daNpc_Jagar_c : public daNpcT_c {
...
+
     static char* mCutNameList[7];
     static cutFunc mCutList[7];
 private:
-    /* 0x0E40 */ int field_0xe40;
+#if DEBUG
+    /* 0x0E40 */ daNpc_Jagar_HIO_c* mpHIO;
+#else
+    /* 0x0E40 */ daNpc_Jagar_Param_c* mpHIO;
+#endif
```
TakaRikka: i wonder if something like this should be handled with a define or macro. i feel like the og devs must've had something like that going instead of just writing ifdef blocks everywhere for HIO stuff. something like
```cpp
#if DEBUG
#define HIO_CLASS daNpc_Jagar_HIO_c
#else
#define HIO_CLASS daNpc_Jagar_Param_c
#endif
```

include/f_op/f_op_actor_mng.h
```diff
@@ -184,6 +184,7 @@ enum fopAcM_CARRY {
     /* 0x30 */ fopAcM_CARRY_UNK_30 = 0x30,
     /* 0x40 */ fopAcM_CARRY_UNK_40 = 0x40,
     /* 0x80 */ fopAcM_CARRY_CHICKEN = 0x80,
+    /* 0x85 */ fopAcM_CARRY_UNK_85 = 0x85,
```
TakaRikka: this prob shouldnt exist, should probably be `(fopAcM_CARRY_CHICKEN | fopAcM_CARRY_SIDE | fopAcM_CARRY_TYPE_1)`. `fopAcM_CARRY_CHICKEN ` probably isnt a correct name, will fix eventually

src/d/actor/d_a_obj_pumpkin.cpp
```diff
@@ -7,765 +7,941 @@
...
+    mAcchCir.SetWallR(mpHIO->m.width * mPumpkinScale);
+    mAcchCir.SetWallH(mpHIO->m.knee_length * mPumpkinScale);
+    gravity = mpHIO->m.gravity * mPumpkinScale;
+    int sp_0x30 = 0;
+    if (fopAcM_checkCarryNow(this) || fopAcM_checkHawkCarryNow(this) || field_0xbb3) {
+        sp_0x30 = 1;
+    }
+
+    if (field_0xBAE) {
+        //
```
TakaRikka: comment can be removed?

src/d/actor/d_a_obj_pumpkin.cpp
```diff
@@ -7,765 +7,941 @@
...
+        int item_table_no = getItemTableNo();
+        if (item_table_no >= 0 && mItemProcId == 0xFFFFFFFF) {
+            mItemProcId = fopAcM_createItemFromTable(&current.pos, item_table_no, -1, fopAcM_GetRoomNo(this),
+                                                     NULL, 0, NULL, NULL, NULL, NULL);
+        }
+
+        if (mItemProcId == 0xFFFFFFFF || fopAcM_IsExecuting(mItemProcId)) {
```
TakaRikka: use fpcM_ERROR_PROCESS_ID_e when checking actor id's

src/d/actor/d_a_obj_pumpkin.cpp
```diff
@@ -7,765 +7,941 @@
...
+    dComIfGp_particle_levelEmitterOnEventMove(mSmokePrtcl2);
 }
 
 /* 80CB7D98-80CB7E98 0026F8 0100+00 1/1 0/0 0/0 .text            setWaterPrtcl__15daObj_Pumpkin_cFv
  */
 void daObj_Pumpkin_c::setWaterPrtcl() {
-    // NONMATCHING
-}
+    static u16 emttrId[4] = {
+        0x01B8, 0x01B9, 0x01BA, 0x01BB,
```
TakaRikka: particle enums

src/d/actor/d_a_obj_pumpkin.cpp
```diff
@@ -7,765 +7,941 @@
...
+    speed.setall(0.0f);
+    field_0xBB0 = 1;
+    field_0xBAE = 1;
 }
 
 /* 80CB81A4-80CB81C4 002B04 0020+00 1/0 0/0 0/0 .text            daObj_Pumpkin_Create__FPv */
-static void daObj_Pumpkin_Create(void* param_0) {
-    // NONMATCHING
+static int daObj_Pumpkin_Create(void* i_this) {
+    int ret = static_cast<daObj_Pumpkin_c*>(i_this)->create();
```
TakaRikka, line 897: does this actually use a temp?

YunataSavior: per the debug ROM, yes

include/d/actor/d_a_npc_jagar.h
YunataSavior: I agree.

Furthermore, I discovered via the debug ROM that the HIO classes MUST be declared above the actor class in the header file. This makes sense.

PR #2722, opened 2025-10-03
link d_a_obj_hasu2

roeming:
I tried my best to match any conventions I could find, let me know if I missed anything.


src/d/actor/d_a_obj_hasu2.cpp
```diff
@@ -139,180 +73,182 @@ extern actor_process_profile_definition g_profile_Obj_MHasu = {
...
+    MTXCopy(mDoMtx_stack_c::get(), mMtx2);
+    MTXCopy(mDoMtx_stack_c::get(), mMtx);
 }
 
 /* 80C18440-80C184B0 000220 0070+00 1/0 0/0 0/0 .text CreateHeap__Q29nObjMHasu12daObjMHasu_cFv */
-void nObjMHasu::daObjMHasu_c::CreateHeap() {
-    // NONMATCHING
+int nObjMHasu::daObjMHasu_c::CreateHeap() {
+    J3DModelData* model_data = (J3DModelData*)dComIfG_getObjectRes(l_arcName, l_bmdidx);
+    JUT_ASSERT(0xc1, model_data != 0);
```
LagoLunatic: Defines in assertion strings get converted to their original value due to how nested defines work.
```suggestion
    JUT_ASSERT(0xc1, model_data != NULL);
```

YunataSavior: Furthermore, please convert the hex val into decimal for these assert/panic messages.

src/d/actor/d_a_obj_hasu2.cpp
```diff
@@ -6,119 +6,53 @@
...
+const int l_dzbidx = {7};
+const int l_bmdidx = {4};
 
 //
 // Declarations:
 //
-
+inline void nObjMHasu::daObjMHasu_c::setRideFlag(bool flag)
+{
+    this->mRideFlag = flag;
```
TakaRikka: no need for `this->`

include/d/actor/d_a_obj_hasu2.h
```diff
@@ -12,23 +13,35 @@ namespace nObjMHasu {
...
+        /* 80C185A4 */ int Execute(Mtx**);
+        /* 80C18600 */ int Draw();
+        /* 80C186A4 */ int Delete();
         /* 80C186E0 */ void upDownHasu();
         /* 80C18888 */ void updateCount(u8);
         /* 80C188A0 */ void setHasuCount(f32);
-        /* 80C189E0 */ ~daObjMHasu_c();
+        /* 80C189E0 */ virtual ~daObjMHasu_c();
 
+        inline void setRideFlag(bool);
```
TakaRikka: this can just be inlined in the class afaict?

src/d/actor/d_a_obj_hasu2.cpp
```diff
@@ -6,119 +6,53 @@
...
-#pragma push
-#pragma force_active on
-SECTION_DEAD static char const* const stringBase_80C18AA0 = "M_Hasu";
-#pragma pop
 
 /* 80C18AA8-80C18AAC -00001 0004+00 3/3 0/0 0/0 .data            l_arcName */
-SECTION_DATA static void* l_arcName = (void*)&d_a_obj_hasu2__stringBase0;
+char* l_arcName = "M_Hasu";
+
+static void daObjMHasu_create1st(nObjMHasu::daObjMHasu_c* i_this);
```
TakaRikka: atm for consistency we move the profile and methods to the end of the TU and remove these prototypes

src/d/actor/d_a_obj_hasu2.cpp
```diff
@@ -139,180 +73,182 @@ extern actor_process_profile_definition g_profile_Obj_MHasu = {
...
+    mConst2(6.5f),
+    mConst3(5.5f)
+{
 }
 
 /* 80C18340-80C183D0 000120 0090+00 1/1 0/0 0/0 .text create1st__Q29nObjMHasu12daObjMHasu_cFv */
-void nObjMHasu::daObjMHasu_c::create1st() {
-    // NONMATCHING
+int nObjMHasu::daObjMHasu_c::create1st() {
+    cPhs__Step phase = (cPhs__Step)dComIfG_resLoad(static_cast<request_of_phase_process_class*>(this), l_arcName);
```
TakaRikka: no cast required here, just pass `this`

src/d/actor/d_a_obj_hasu2.cpp
```diff
@@ -139,180 +73,182 @@ extern actor_process_profile_definition g_profile_Obj_MHasu = {
...
-};
-
 /* 80C182DC-80C18340 0000BC 0064+00 1/1 0/0 0/0 .text            __ct__Q29nObjMHasu12daObjMHasu_cFv
  */
-nObjMHasu::daObjMHasu_c::daObjMHasu_c() {
-    // NONMATCHING
+nObjMHasu::daObjMHasu_c::daObjMHasu_c() :
+    mConst1(25.5f),
+    mConst2(6.5f),
+    mConst3(5.5f)
```
TakaRikka, line 43: we prefer to set these directly rather than use initializers
```cpp
nObjMHasu::daObjMHasu_c::daObjMHasu_c() {
    mConst1 = 25.5f;
    // ...
}
```

src/d/actor/d_a_obj_hasu2.cpp
```diff
@@ -139,180 +73,182 @@ extern actor_process_profile_definition g_profile_Obj_MHasu = {
...
+    }
+    unk60c_1 = mRideFlag;
+    mRideFlag = 0;
 }
 
 /* 80C18888-80C188A0 000668 0018+00 1/1 0/0 0/0 .text updateCount__Q29nObjMHasu12daObjMHasu_cFUc
  */
-void nObjMHasu::daObjMHasu_c::updateCount(u8 param_0) {
-    // NONMATCHING
+void nObjMHasu::daObjMHasu_c::updateCount(u8 count) {
```
TakaRikka: parameters named like `i_count` for consistency unless an assert says otherwise

src/d/actor/d_a_obj_hasu2.cpp
```diff
@@ -139,180 +73,182 @@ extern actor_process_profile_definition g_profile_Obj_MHasu = {
...
+    }
 }
 
 /* 80C18920-80C18974 000700 0054+00 1/0 0/0 0/0 .text
  * daObjMHasu_create1st__FPQ29nObjMHasu12daObjMHasu_c           */
-static void daObjMHasu_create1st(nObjMHasu::daObjMHasu_c* param_0) {
-    // NONMATCHING
+static void daObjMHasu_create1st(nObjMHasu::daObjMHasu_c* i_this) {
+    fopAcM_SetupActor(i_this, nObjMHasu::daObjMHasu_c);
+    i_this->create1st();
```
TakaRikka: all actor methods (create, delete, isdelete, draw, execute) return int, returning the inner call

src/d/actor/d_a_obj_hasu2.cpp
```diff
@@ -139,180 +73,182 @@ extern actor_process_profile_definition g_profile_Obj_MHasu = {
...
-    // NONMATCHING
+static void daObjMHasu_MoveBGDraw(nObjMHasu::daObjMHasu_c* i_this) {
+    i_this->MoveBGDraw();
 }
 
 /* 80C189E0-80C18A50 0007C0 0070+00 1/0 0/0 0/0 .text            __dt__Q29nObjMHasu12daObjMHasu_cFv
  */
-nObjMHasu::daObjMHasu_c::~daObjMHasu_c() {
-    // NONMATCHING
+nObjMHasu::daObjMHasu_c::~daObjMHasu_c()
```
TakaRikka: this is probably compiler generated

include/d/actor/d_a_obj_hasu2.h
```diff
@@ -12,23 +13,35 @@ namespace nObjMHasu {
...
 private:
-    /* 0x568 */ u8 field_0x568[0x634 - 0x568];
+    /* 0x5A8 */ Mtx mMtx;
+    /* 0x5D8 */ Mtx mMtx2;
+    /* 0x608 */ J3DModel* mModel;
+    /* 0x60C */ bool mRideFlag : 1;
+    /* 0x60C */ bool unk60c_1 : 1;
```
TakaRikka, line 37: are these actually bitfields? im skeptical since basically nowhere outside of Z2Audio uses bitfield members

src/d/actor/d_a_obj_hasu2.cpp
```diff
@@ -6,119 +6,53 @@
...
+{
+    this->mRideFlag = flag;
+}
 /* 80C18298-80C182DC 000078 0044+00 1/1 0/0 0/0 .text
  * rideCallBack__FP4dBgWP10fopAc_ac_cP10fopAc_ac_c              */
 static void rideCallBack(dBgW* param_0, fopAc_ac_c* param_1, fopAc_ac_c* param_2) {
-    // NONMATCHING
+    (void)param_0;
+    if (fopAcM_GetName(param_2) == PROC_ALINK) {
+        nObjMHasu::daObjMHasu_c* hasu = static_cast<nObjMHasu::daObjMHasu_c*>(param_1);
```
TakaRikka: C style casts preferred over C++ casts

src/d/actor/d_a_obj_hasu2.cpp
```diff
@@ -139,180 +73,182 @@ extern actor_process_profile_definition g_profile_Obj_MHasu = {
...
-    // NONMATCHING
+    { // nested scope from debug
+        f32 step;
+        f32 target;
+        if (mRideFlag) {
+            target = unk62C + (mPos.y - mConst1);
+            if (current.pos.y > (mPos.y - mConst1) + mConst2 + mConst2) {
+                step = mStep + mConst3;
+            } else if (mTimer) {
+                step = 6.f * mStep;
```
TakaRikka: zeroes preferred for whole floats for consistency (eg 6.0f)

include/d/actor/d_a_obj_hasu2.h
roeming: It ends up not linking if non-virtual. The vtable ends up not matching.

src/d/actor/d_a_obj_hasu2.cpp
roeming: This method doesn't match debug

src/d/actor/d_a_obj_hasu2.cpp
roeming: removing it breaks linking, it seems to need a virtual dtor

include/d/actor/d_a_obj_hasu2.h
roeming: Not totally sure, but this matches debug

src/d/actor/d_a_obj_hasu2.cpp
roeming: Is the convention just to use `i_` for all parameters?

include/d/actor/d_a_obj_hasu2.h
LagoLunatic: You don't need to make it virtual to inline a function, that's unrelated. Just put it inside the class body and then remove it from the cpp file. This is how most inlines are defined, excluding very large inlines.
```suggestion
        void setRideFlag(bool flag) { mRideFlag = flag; }
```

src/d/actor/d_a_obj_hasu2.cpp
LagoLunatic: Instead of removing it, have you tried putting `virtual ~daObjMHasu_c() {}` in the header instead of the cpp?

hatal175 (approved):

PR #2728, opened 2025-10-07
lv4bridge / lv4prwall done

TakaRikka:


YunataSavior:
Looks like someone with JPN/PAL disc will have to match lv4bridge

PR #2734, opened 2025-10-10
d_a_e_sf matching

carter-ktb21:


include/d/actor/d_a_e_sf.h
```diff
@@ -9,18 +12,89 @@
...
+    /* 0x68A */ s16 mAction;
+    /* 0x68C */ f32 mPlayerDistanceXZ;
+    /* 0x690 */ s16 mPlayerAngleY;
+    /* 0x692 */ s16 mPlayerAngleX;
+    /* 0x694 */ f32 mRecognizeDist;
+    /* 0x698 */ u32 mShadowKey;
+    /* 0x69C */ s16 mTimers[4];
+    /* 0x6A4 */ s16 mInvulnerabilityTimer;
+    /* 0x6A6 */ s16 mColor;
+    /* 0x6A8 */ u8 unused_u8;
```
hatal175: Let's stick with the field_* convention for the unused fields.

src/d/actor/d_a_e_sf.cpp
```diff
@@ -7,1352 +7,1931 @@
...
+            model->getModelData()->getJointNodePointer(i)->setCallBack(nodeCallBack);
+        }
+    }
+
+    static int wepon_data[2] = {
+        BMDR_SF_SWORDA, BMDR_SF_SWORDB,
+    };
+
+    J3DModelData* modelData = (J3DModelData*)dComIfG_getObjectRes("E_sf", wepon_data[i_this->mType]);
+    JUT_ASSERT(2752, modelData != 0);
```
hatal175: You can use != NULL

src/d/actor/d_a_e_sf.cpp
```diff
@@ -7,1352 +7,1931 @@
...
+    if (i_this->mSwordModel == NULL) {
+        return 0;
+    }
+
+    static int tate_data[2] = {
+        BMDR_SF_SHIELDA, BMDR_SF_SHIELDB,
+    };
+
+    modelData = (J3DModelData*)dComIfG_getObjectRes("E_sf", tate_data[i_this->arg3]);
+    JUT_ASSERT(2762, modelData != 0);
```
hatal175: You can use != NULL

src/d/actor/d_a_e_sf.cpp
YunataSavior: Huh, I guess I need to start doing this too

PR #2736, opened 2025-10-14
Enemy Variable Cleanup

carter-ktb21:


include/d/actor/d_a_b_bh.h
```diff
@@ -28,17 +29,17 @@ class b_bh_class : public fopEn_enemy_c {
     /* 0x664 */ u8 field_0x664[0x668 - 0x664];
     /* 0x668 */ s16 field_0x668;
     /* 0x66A */ s16 mAction;
-    /* 0x66C */ s16 mMode;
+    /* 0x66C */ s16 mActionPhase;
```
Luzagar, line 32: Is there any reason to use mActionPhase instead of mMode or mode Because on some insert we have the  devs use mMoveMode or movemode

include/d/actor/d_a_b_gnd.h
```diff
@@ -60,7 +60,7 @@ class b_gnd_class : public fopEn_enemy_c {
     /* 0x0772 */ s16 field_0x772;
     /* 0x0774 */ dKy_tevstr_c mSwordTevstr;
     /* 0x0AFC */ s16 mCounter;
-    /* 0x0AFE */ s16 mActionMode;
+    /* 0x0AFE */ s16 mAction;
```
Luzagar, line 63: Why change it ? There are both official name in assert

include/d/actor/d_a_b_bh.h
carco-games: I saw actionPhase in d_a_e_yk, but if there are debug asserts, than I will rework the phase naming

carco-games: I'm actually not sure I see any asserts specifying "mMode". Could you point to and instance of it?

Luzagar: my bad on mMode but it seem those TU who use movemode or mMoveMode seem to don't use anm_init so yeah

TakaRikka: these c style actors need to have their member names changed to c style names anyway which i was just saving for a later cleanup pass since its not a priority. the current name doesnt matter too much, will change it later

include/d/actor/d_a_b_bq.h
```diff
@@ -35,25 +35,25 @@ class b_bq_class : public fopEn_enemy_c {
     /* 0x06C4 */ int mAnmID;
     /* 0x06C8 */ s16 field_0x6c8;
     /* 0x06CA */ s16 mAction;
-    /* 0x06CC */ s16 mMode;
+    /* 0x06CC */ s16 mActionPhase;
```
TakaRikka, line 38: as a follow up to the last comment, i'd probably just refrain from changing these names unless there's an assert within the TU its used in since its just gonna be changed later anyway. giving names to unknown fields is good, but these already named ones dont really need changes unless its wildly inaccurate

PR #2744, opened 2025-10-18
m_Do_main / f_ap_game / f_op_actor debug work and misc

TakaRikka:


src/f_op/f_op_actor.cpp
```diff
@@ -34,194 +218,355 @@ u32 fopAc_ac_c::stopStatus;
...
+            OSReport("<%s> の処理を停止します\n", l_name);
+        }
+
+        l_HIO.field_0x7 = l_HIO.mStopExecute;
+        return 1;
+    }
+
+    l_HIO.field_0x7 = l_HIO.mStopExecute;
+
+    JUT_ASSERT(662, !(fpclassify(actor->current.pos.x) == 1));
```
Luzagar, line 311: does that confirm that the official name of a_this is actor in c style TU ?

TakaRikka: wouldn't say confirms but generally its the style ive been going for now yes

PR #2754, opened 2025-10-23
link d_a_obj_lv6elevta

roeming:
The only thing I was unsure of was the explicit construction of `JAISoundID` for functions like `Z2GetAudioMgr()->seStart`, so I left them in. They also match if you just pass in the enum and have the object implicitly constructed.

### DIFF
--- a/src/d/actor/d_a_obj_lv6elevta.cpp
+++ b/src/d/actor/d_a_obj_lv6elevta.cpp
@@ -139,8 +139,8 @@ void daObjLv6ElevtA_c::moveAngle(void* i_FoundActor) {
     }
     shape_angle = current.angle;
     mIsMoving = true;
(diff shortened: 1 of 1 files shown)


src/d/actor/d_a_obj_lv6elevta.cpp
```diff
@@ -1,315 +1,314 @@
...
+    } else {
+        f32 v = (mAngle / (f32)0x3fff) * 800.0f;
+        current.pos.y = home.pos.y - v;
+        current.angle.y = home.angle.y + mAngle;
+        found->current.pos.y = found->home.pos.y - v;
+        found->current.angle.y = found->home.angle.y + mAngle;
+    }
+    shape_angle = current.angle;
+    mIsMoving = true;
+    Z2GetAudioMgr()->seStartLevel(JAISoundID(Z2SE_OBJ_L6_RL_FLR_MV), &current.pos, 0, 0, 1.0f, 1.0f,
```
TakaRikka: yeah sound ids are implicitly constructed so you can remove the explicit JAISoundID's

PR #2757, opened 2025-10-25
NPC work, obj_szbridge matching

hev7:


src/d/actor/d_a_npc_ashB.cpp
```diff
@@ -1109,42 +982,42 @@ BOOL daNpcAshB_c::EvCut_Appear(int i_staffID) {
     }
 
     switch (*cut_name) {
-        case 0x30303031:
-        case 0x30303032:
+    case 0x30303031:
```
hatal175: Mind fixing these to literals if you're here?

PR #2762, opened 2025-10-30
d_a_npc_bans almost matching

carter-ktb21:
Two tiny issues in two functions are preventing matches. If anyone has any ideas for resolving, please share!

### DIFF
--- a/include/d/d_shop_camera.h
+++ b/include/d/d_shop_camera.h
@@ -8,7 +8,7 @@ class ShopCam_action_c {
 public:
     typedef int (ShopCam_action_c::*ShopCamFunc)();
 
-    inline ShopCam_action_c() {
+    ShopCam_action_c() {
         setCameraSpeed(0.25f, 20.0f, 0.25f, 5.0f);
         field_0xda = true;
         field_0xd0 = 0;
--- a/src/d/actor/d_a_npc_bans.cpp
+++ b/src/d/actor/d_a_npc_bans.cpp
@@ -457,11 +457,6 @@ cPhs__Step daNpc_Bans_c::create() {
     return phase;
 }
 
-enum unk_ModelFlags_e {
-    unk_flag_1 = 0,
-    unk_flag_2 = 0x80000,
-};
-
 /* 80962D6C-8096317C 00058C 0410+00 1/1 0/0 0/0 .text            CreateHeap__12daNpc_Bans_cFv */
 int daNpc_Bans_c::CreateHeap() {
     J3DModelData* modelData;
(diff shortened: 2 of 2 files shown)


include/d/d_shop_camera.h
```diff
@@ -4,9 +4,29 @@
 #include "f_op/f_op_actor.h"
 
 
-struct ShopCam_action_c {
+class ShopCam_action_c {
+public:
     typedef int (ShopCam_action_c::*ShopCamFunc)();
 
+    inline ShopCam_action_c() {
```
TakaRikka: inline keyword should be able to be removed

src/d/actor/d_a_npc_bans.cpp
```diff
@@ -521,1202 +310,1653 @@ char* daNpc_Bans_c::mCutNameList[4] = {
...
+        Execute();
+        mCreating = false;
+    }
+
+    return phase;
+}
+
+enum unk_ModelFlags_e {
+    unk_flag_1 = 0,
+    unk_flag_2 = 0x80000,
```
TakaRikka: the actual flags should be in J3DModel.h iirc, no need to use unk flags

src/d/actor/d_a_npc_bans.cpp
```diff
@@ -521,1202 +310,1653 @@ char* daNpc_Bans_c::mCutNameList[4] = {
...
+    }
+
+    selectAction();
+    srchActors();
+
+    #if DEBUG
+    u32 flags = 0x42;
+    #else
+    u32 flags = 10;
+    #endif
```
TakaRikka: use fopAc_attnflag enums, which should also remove the need for using an ifdef block

src/d/actor/d_a_npc_bans.cpp
```diff
@@ -521,1202 +310,1653 @@ char* daNpc_Bans_c::mCutNameList[4] = {
...
+
+    return NULL;
+}
 
 /* 80964F48-80964FEC 002768 00A4+00 1/1 0/0 0/0 .text            setPrtcls__12daNpc_Bans_cFv */
 void daNpc_Bans_c::setPrtcls() {
-    // NONMATCHING
-}
+    static u16 const id[4] = {
+        0x862C, 0x86A3, 0x862D, 0x862E,
```
TakaRikka: particle name enums

src/d/actor/d_a_npc_bans.cpp
```diff
@@ -521,1202 +310,1653 @@ char* daNpc_Bans_c::mCutNameList[4] = {
...
+
+        case 3:
+            break;
+    }
+
+    return 0;
+}
+
+#if DEBUG
+/* 80DA60D0 - 80DA61AF */
```
TakaRikka: you can leave out the address comments since after GC USA is done i want to remove all of them anyway so that its more version agnostic

src/d/actor/d_a_npc_bans.cpp
```diff
@@ -521,1202 +310,1653 @@ char* daNpc_Bans_c::mCutNameList[4] = {
...
+    }
+
+    return rv;
+}
+#endif
 
 /* 80966C88-80966CA8 0044A8 0020+00 1/0 0/0 0/0 .text            daNpc_Bans_Create__FPv */
-static void daNpc_Bans_Create(void* param_0) {
-    // NONMATCHING
+static int daNpc_Bans_Create(void* a_this) {
```
TakaRikka: i_this for these

TakaRikka (approved):

PR #2766, opened 2025-11-01
d_a_npc_bous equivalent

carter-ktb21:
Weak func issue I'm not sure how to fix


src/d/actor/d_a_npc_bouS.cpp
```diff
@@ -311,1617 +222,1455 @@ static char* l_evtNames[4] = {
...
-    // NONMATCHING
+    s16 attention_distance = 5;
+    s16 attention_angle = 6;
+    s16 talk_distance = 3;
+    s16 talk_angle = 6;
+
+    attention_info.distances[fopAc_attn_LOCK_e] = getDistTableIdx(attention_distance, attention_angle);
+    attention_info.distances[fopAc_attn_TALK_e] = attention_info.distances[fopAc_attn_LOCK_e];
+    attention_info.distances[fopAc_attn_SPEAK_e] = getDistTableIdx(talk_distance, talk_angle);
+    #if DEBUG
```
TakaRikka: i get the feeling this isnt a debug thing but rather a version difference, so i'd prob use `PLATFORM_SHIELD` here instead

PR #2770, opened 2025-11-04
d_a_e_yg matching

carter-ktb21:


src/d/actor/d_a_e_yg.cpp
```diff
@@ -7,1470 +7,1493 @@
...
 }
 
 /* 807FA2E4-807FA478 002084 0194+00 1/1 0/0 0/0 .text            damage_check__FP10e_yg_class */
-static void damage_check(e_yg_class* param_0) {
-    // NONMATCHING
+static void damage_check(e_yg_class* i_this) {
+    fopAc_ac_c* actor = (fopAc_ac_c*)&i_this->actor;
+    daPy_py_c* player = (daPy_py_c*)dComIfGp_getPlayer(0);
+
+    if (!i_this->mInvulnerabilityTimer) {
```
TakaRikka: since this is a timer, compare to 0 explicitly

PR #2772, opened 2025-11-05
d_a_e_sh matching

caseif:
`d_a_e_sh` is now matching with this PR. Additionally, I've re-added `#pragma pack()` directives where they were previously removed as my IDE (CLion) doesn't handle push/pop properly for packing which messes with its struct analysis. They should effectively be no-ops in practice since they're immediately followed by `#pragma pop`.

### DIFF
--- a/src/d/actor/d_a_e_sh.cpp
+++ b/src/d/actor/d_a_e_sh.cpp
@@ -9,1027 +9,1299 @@
 #include "d/d_cc_d.h"
 #include "dol2asm.h"
 #include "f_op/f_op_actor_enemy.h"
+#include "f_op/f_op_camera_mng.h"
 
+class daE_SH_HIO_c : public JORReflexible {
+public:
+    /* 8078E34C */ daE_SH_HIO_c();
+    /* 807918B4 */ virtual ~daE_SH_HIO_c() {}
+#if DEBUG
+    void genMessage(JORMContext*);
+#endif
 
-//
-// Forward References:
-//
-
-extern "C" void __ct__12daE_SH_HIO_cFv();
-extern "C" static void anm_init__FP10e_sh_classifUcf();
-extern "C" static void nodeCallBack__FP8J3DJointi();
-extern "C" static void daE_SH_Draw__FP10e_sh_class();
-extern "C" static void damage_check__FP10e_sh_class();
-extern "C" static void e_sh_stop__FP10e_sh_class();
-extern "C" static void e_sh_appear__FP10e_sh_class();
-extern "C" static void e_sh_move__FP10e_sh_class();
-extern "C" static void e_sh_attack__FP10e_sh_class();
-extern "C" static void e_sh_disappear__FP10e_sh_class();
-extern "C" static void e_sh_damage__FP10e_sh_class();
-extern "C" static void e_sh_dead__FP10e_sh_class();
-extern "C" static void action__FP10e_sh_class();
-extern "C" static void eff_set__FP10e_sh_class();
-extern "C" static void anm_se_set__FP10e_sh_class();
-extern "C" static void daE_SH_Execute__FP10e_sh_class();
-extern "C" static bool daE_SH_IsDelete__FP10e_sh_class();
-extern "C" static void daE_SH_Delete__FP10e_sh_class();
-extern "C" static void useHeapInit__FP10fopAc_ac_c();
-extern "C" static void daE_SH_Create__FP10fopAc_ac_c();
-extern "C" void __dt__8cM3dGSphFv();
-extern "C" void __dt__8cM3dGAabFv();
-extern "C" void __dt__8dCcD_SphFv();
-extern "C" void __ct__8dCcD_SphFv();
-extern "C" void __dt__10dCcD_GSttsFv();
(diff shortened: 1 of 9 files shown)


src/d/actor/d_a_e_sh.cpp
```diff
@@ -9,1027 +9,1299 @@
...
-  0x00040100,             // mStatus
-  fopAc_ENEMY_e,          // mActorType
-  fopAc_CULLBOX_CUSTOM_e, // cullType
-};
+/* 8078EA58-8078EC04 0007F8 01AC+00 1/1 0/0 0/0 .text            e_sh_stop__FP10e_sh_class */
+static void e_sh_stop(e_sh_class* i_this) {
+    static u16 ap_name[3] = {
+        0x81CD,
+        0x81CE,
+        0x81CF,
```
TakaRikka: particle enums

src/d/actor/d_a_e_sh.cpp
```diff
@@ -9,1027 +9,1299 @@
...
+    };
+
+    fopAc_ac_c* actor = (fopAc_ac_c*)&i_this->enemy;
+
+    i_this->field_0x6a0 = 6;
+
+    dScnKy_env_light_c* envLight = dKy_getEnvlight();
+    int unkInt1 = envLight->daytime / 15.0f;
```
TakaRikka: this should be the hour component of the time of day, which makes sense since stalhounds only appear at night

src/d/actor/d_a_e_sh.cpp
```diff
@@ -9,1027 +9,1299 @@
...
 }
 
 /* 8078FA38-8078FBC8 0017D8 0190+00 1/1 0/0 0/0 .text            e_sh_disappear__FP10e_sh_class */
-static void e_sh_disappear(e_sh_class* param_0) {
-    // NONMATCHING
+static void e_sh_disappear(e_sh_class* i_this) {
+    static u16 dp_name[3] = {
+        0x81D0,
+        0x81D1,
+        0x81D2,
```
TakaRikka: particle enums

src/d/actor/d_a_e_sh.cpp
```diff
@@ -9,1027 +9,1299 @@
...
+/* 80790278-80790498 002018 0220+00 1/1 0/0 0/0 .text            eff_set__FP10e_sh_class */
+static void eff_set(e_sh_class* i_this) {
+    static u16 p_name[2] = {
+        0x81D4,
+        0x81D5,
+    };
+    static u32 p_idx[2] = {
+        0xD,
+        0xE,
+    };
```
TakaRikka, line 875: particle enums, and p_idx should prob be int's in decimal

src/d/actor/d_a_e_sh.cpp
```diff
@@ -9,1027 +9,1299 @@
...
 
-/* 807915E0-807916AC 003380 00CC+00 1/1 0/0 0/0 .text            __dt__8dCcD_SphFv */
-// dCcD_Sph::~dCcD_Sph() {
-extern "C" void __dt__8dCcD_SphFv() {
-    // NONMATCHING
-}
+/* 80790FC8-80791030 002D68 0068+00 1/0 0/0 0/0 .text            daE_SH_Delete__FP10e_sh_class */
+static int daE_SH_Delete(e_sh_class* i_this) {
+    fopAc_ac_c* actor = (fopAc_ac_c*)&i_this->enemy;
+    int unusedId = fopAcM_GetID(i_this);
```
TakaRikka: fpc_ProcID

src/d/actor/d_a_e_sh.cpp
```diff
@@ -9,1027 +9,1299 @@
...
+
+    sh->mAnm_p = new mDoExt_McaMorfSO((J3DModelData*)dComIfG_getObjectRes("E_sh", 22), NULL, NULL,
+                                      (J3DAnmTransform*)dComIfG_getObjectRes("E_sh", 14), 2, 1.0f,
+                                      0, -1, &sh->mSound, 0x80000, 0x11000084);
+
+    if (sh->mAnm_p == NULL || sh->mAnm_p->getModel() == NULL) {
+        return 0;
+    } else {
+        J3DModel* model = sh->mAnm_p->getModel();
+        model->setUserArea((u32)sh);
```
TakaRikka: these days i prefer to use `uintptr_t` for pointers being treated as uint/u32

src/d/actor/d_a_e_sh.cpp
```diff
@@ -9,1027 +9,1299 @@
...
+            {0x0}, // mGObjCo
+        }, // mObjInf
+        {
+            {{0.0f, 0.0f, 0.0f}, 40.0f} // mSph
+        } // mSphAttr
+    };
+
+    e_sh_class* sh = (e_sh_class*)i_this;
+
+    fopAcM_ct((fopAc_ac_c*)sh, e_sh_class);
```
TakaRikka, line 1209: is this not using the in parameter?

caseif: Apparently not, according to debug.

PR #2797, opened 2025-11-11
rename variable to match assert

roeming:
`daNpcF_c::mpMorf` -> `daNpcF_c::mAnm_p`, found in `d_a_npc_gra`


roeming:
I'm fine to close this. I noticed that the other pr that was merged renamed a lot more fields than just the one referenced in the assert. I know we have very few asserts to actually know what the fields are called, but just raising that the previous pr renamed fields from other classes that shared a name with the npcf field.

PR #2799, opened 2025-11-13
d_a_npc_theb matching

carter-ktb21:


src/d/actor/d_a_npc_theB.cpp
```diff
@@ -307,1166 +146,1503 @@ static char* l_arcName = "Coach";
...
+        mLookat.setAttnPos(NULL);
+    }
+
+    mLookat.setParam(body_angleX_min, body_angleX_max, body_angleY_min, body_angleY_max, 0.0f, 0.0f,
+                     0.0f, 0.0f, head_angleX_min, head_angleX_max, head_angleY_min, head_angleY_max,
+                     mCurAngle.y, lookat_pos);
+    mLookat.calc(this, model->getBaseTRMtx(), lookat_angle, FALSE, angle_delta, false);
+}
+
+#pragma inline_max_size(503) // FAKEMATCH
```
YunataSavior: You may want to write a comment here linking this back to the DANPCF HACK flag in configure.py and how, once we find the proper solution, we must remove this line.

TakaRikka (approved):

PR #2802, opened 2025-11-14
npc_kn almost matching + dbg work

YunataSavior:
I give up on `daNpc_Kn_c::setSwordChargePtcl`...

### DIFF
--- a/include/Z2AudioLib/Z2AudioMgr.h
+++ b/include/Z2AudioLib/Z2AudioMgr.h
@@ -17,13 +17,7 @@
 class JKRArchive;
 class JKRSolidHeap;
 
-class UnkZ2Dbg {
-#if DEBUG
-    u8 pad[0x14];
-#endif
-};
-
-class Z2AudioMgr : public Z2SeMgr, public UnkZ2Dbg, public Z2SeqMgr, public Z2SceneMgr, public Z2StatusMgr, public Z2SoundObjMgr {
+class Z2AudioMgr : public Z2SeMgr, public Z2SeqMgr, public Z2SceneMgr, public Z2StatusMgr, public Z2SoundObjMgr {
 public:
     Z2AudioMgr();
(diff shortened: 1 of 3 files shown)


src/d/actor/d_a_npc_kn.cpp
```diff
@@ -22,6 +26,139 @@ const dCcD_SrcGObjInf daNpc_Kn_c::mCcDObjData = {
...
+    2.0f,
+};
+
+#if DEBUG
+daNpc_Kn_HIO_c::daNpc_Kn_HIO_c() {
+    m = daNpc_Kn_Param_c::m;
+}
+
+void daNpc_Kn_HIO_c::listenPropertyEvent(const JORPropertyEvent* event) {
+    char msg_buffer[1988];
```
hatal175: does 2000 fit? then you can use sizeof in the memset call

include/Z2AudioLib/Z2AudioMgr.h
```diff
@@ -17,7 +17,13 @@
...
 class JKRSolidHeap;
 
-class Z2AudioMgr : public Z2SeMgr, public Z2SeqMgr, public Z2SceneMgr, public Z2StatusMgr, public Z2SoundObjMgr {
+class UnkZ2Dbg {
+#if DEBUG
+    u8 pad[0x14];
+#endif
+};
+
+class Z2AudioMgr : public Z2SeMgr, public UnkZ2Dbg, public Z2SeqMgr, public Z2SceneMgr, public Z2StatusMgr, public Z2SoundObjMgr {
```
TakaRikka: idk this seems wrong? we have RTTI so we know what the inheritance structure is supposed to look like. could this be extra members of Z2SeMgr?

YunataSavior: Done

TakaRikka (approved):

PR #2806, opened 2025-11-14
d_a_npc_pouya matching

hev7:


src/d/actor/d_a_npc_pouya.cpp
```diff
@@ -487,1131 +164,1111 @@ char* daNpc_Pouya_c::mCutNameList[3] = {
...
+        mJntAnm.lookNone(0);
+        local_58[0] = local_60;
+        if (talkProc(local_58, 0, NULL, 0)) {
+            if (local_60 == 0) {
+                int local_64;
+                int evt_id = mFlow.getEventId(&local_64);
+                field_0xfce = 0;
+                switch (evt_id) {
+                case 1:
+                    if (mItemPartnerId == -1) {
```
TakaRikka: mItemPartnerId should be a fpc_ProcID, so use fpcM_ERROR_PROCESS_ID_e

TakaRikka (approved):

PR #2815, opened 2025-11-16
wii building OK / m_Do_graphic debug work

TakaRikka:


src/d/actor/d_a_obj_rstair.cpp
```diff
@@ -346,7 +346,9 @@ int daObjRotStair_c::Draw() {
                 Mtx lightMtx;
                 C_MTXLightPerspective(lightMtx, dComIfGd_getView()->fovy,
                                       dComIfGd_getView()->aspect, 1.0f, 1.0f, -0.01f, 0);
+                #if WIDESCREEN_SUPPORT
```
YunataSavior, line 349: Seems like something we might need to make a macro for?

hatal175 (approved):

PR #2822, opened 2025-11-18
d_a_npc_chin matching

caseif:
`d_a_npc_chin` is matching with this PR. There are also some incidental changes to prevent regressions in debug as a result of changing the signature of `dMsgFlow_c::getChoiceNo` - one regression is still reported, but it's essentially noise due to missing debug code.

### DIFF
--- a/src/d/actor/d_a_npc_chin.cpp
+++ b/src/d/actor/d_a_npc_chin.cpp
@@ -6,400 +6,111 @@
 #include "d/dolzel_rel.h" // IWYU pragma: keep
 
 #include "d/actor/d_a_npc_chin.h"
-#include "dol2asm.h"
-
-//
-// Forward References:
-//
-
-extern "C" void loadModel__Q211daNpcChin_c12_SpotLight_cFv();
-extern "C" void setMtx__Q211daNpcChin_c12_SpotLight_cFv();
-extern "C" void draw__Q211daNpcChin_c12_SpotLight_cFP12dKy_tevstr_c();
-extern "C" void __ct__11daNpcChin_cFv();
-extern "C" void __dt__11daNpcChin_cFv();
-extern "C" void Create__11daNpcChin_cFv();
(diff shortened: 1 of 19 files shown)


src/d/actor/d_a_npc_chin.cpp
```diff
@@ -461,1961 +141,2106 @@ static char* l_evtNames[8] = {
...
 
-/* ############################################################################################## */
-/* 809919A4-809919A8 0000AC 0004+00 18/18 0/0 0/0 .rodata          @4627 */
-SECTION_RODATA static f32 const lit_4627 = -1.0f;
-COMPILER_STRIP_GATE(0x809919A4, &lit_4627);
+    if (dComIfGp_event_runCheck() != FALSE && eventInfo.checkCommandTalk() == FALSE &&
+        field_0xdf4 != 0xffffffff)
+    {
+        dComIfGp_event_setItemPartnerId(field_0xdf4);
+        field_0xdf4 = 0xffffffff;
```
TakaRikka: looks like a fpc_ProcID so you can use fpcM_ERROR_PROCESS_ID_e

src/d/actor/d_a_npc_chin.cpp
```diff
@@ -461,1961 +141,2106 @@ static char* l_evtNames[8] = {
...
+            u16 eventId = mFlow.getEventId(&itemId1);
+            if (eventId != 1) {
+                itemId1 = 0;
+            }
+
+            fpc_ProcID itemId2 = fopAcM_createItemForPresentDemo(&current.pos, itemId1, 0,
+                                                                0xffffffff, 0xffffffff, 0, 0);
+            if (itemId2 != -1) {
+                dComIfGp_event_setItemPartnerId(itemId2);
+            }
```
TakaRikka: fpcM_ERROR_PROCESS_ID_e here too

TakaRikka (approved):

PR #2826, opened 2025-11-19
d_a_kytag13 matching + small various work

carter-ktb21:


TakaRikka:
overall looks fine but may want to check that obj_ice_s change since it broke a wii match

PR #2830, opened 2025-11-19
Misc fixes

LagoLunatic:


YunataSavior:
@TakaRikka I think lago's method is probably more accurate. And I prefer it much more than what I did.

LagoLunatic:
> im curious whether this macro is more accurate than potentially having the wrong inlining flag, but maybe will need to dive into other versions to know for sure

Using `inline_max_size` in configure.py affects every function in the entire TU, which breaks a lot of things since the flag is wrong. The pikmin macro I stole only affects functions that it's used in so it doesn't have side effects outside of the ctor itself.

PR #2834, opened 2025-11-20
add debug functionality to d_a_horse, match almost all functions

roeming:


roeming:
looking into the wii funcs

PR #2836, opened 2025-11-20
Miscellaneous retail+debug fixes, link several new TUs

caseif:
This PR includes a whole bunch of miscellaneous improvements for both retail and debug and links several new TUs for retail.

- Link `d_a_mant`, `d_a_obj_lv7PropellerY`, `d_a_obj_lv7bridge`, and `d_a_e_ws`
- `d_a_npc_cdn3` is now equivalent in retail (`csXyz` destructor is still misplaced) and close to matching in debug (most remaining issues are related to `.data` pooling)
- Link `d_a_midna` and `d_a_npc_shoe` for PAL and JP
- Link `d_a_mg_fshop` for JP
- Small fixes for `J3DSkinDeform` and `m_Do_ext`
- Debug fixes for `d_a_obj_cblock` and `d_a_obj_ss_item`


src/JSystem/J3DGraphAnimator/J3DSkinDeform.cpp
```diff
@@ -293,19 +293,33 @@ int J3DSkinDeform::initMtxIndexArray(J3DModelData* pModelData) {
...
+                pDListPos += r23 * uVar13
+#if PLATFORM_GCN
+                , pDListPos += 3
+#endif
+                )
             {
                 u8 command = *pDListPos;
+#if !PLATFORM_GCN
                 pDListPos++;
+#endif
```
TakaRikka, line 310: i feel like a fakematch comment might be appropriate for these changes, since these dont feel much like actual version changes

src/d/actor/d_a_obj_cblock.cpp
```diff
@@ -51,12 +76,59 @@ static dCcD_SrcCyl l_cyl_src = {
...
+    mTravelDist = 50.0f;
+    mTravelTime = 13;
+    mWaitTime = 8;
+    field_0x14 = 35.0f;
+    field_0x18 = 35.0f;
+#if DEBUG
+    field_0x1c = 15.0f;
+#else
+    field_0x1c = 35.0f;
+#endif
```
TakaRikka: seems more like a version difference rather than debug?

caseif:
Updated to fix small conflict with #2837, both PRs should now apply cleanly onto each other.

include/d/actor/d_a_npc_cdn3.h
```diff
@@ -102,32 +103,47 @@ class daNpcCdn3_c : public daNpcCd2_c {
...
+        return day;
     }
 
+#if PLATFORM_SHIELD
+    void setSpeed(f32 param_2, f32 param_3, f32* param_1, int param_4) {
+#define NPC_CDN3_SET_SPEED(npc_p, param_1, param_2, param_3, param_4) npc_p->setSpeed(param_2, param_3, param_1, param_4)
+#else
     void setSpeed(f32* param_1, f32 param_2, f32 param_3, int param_4) {
+#define NPC_CDN3_SET_SPEED(npc_p, param_1, param_2, param_3, param_4) npc_p->setSpeed(param_1, param_2, param_3, param_4)
+#endif
```
TakaRikka: this kinda feels like a fakematch? seems like a very weird change

include/d/actor/d_a_tag_schedule.h
```diff
@@ -11,13 +11,17 @@ class daTagSchedule_c : public fopAc_ac_c {
...
-    u16 getStartTime() { return (fopAcM_GetParam(this) >> 12) & 0xff; }
+    u8 getStartTime() { return (fopAcM_GetParam(this) >> 12) & 0xff; }
     u8 getStartEnd() { return (fopAcM_GetParam(this) >> 20) & 0xff; }
     int getGroupID() { return fopAcM_GetParam(this) & 0x3f; }
     int getWeekNum() { return mWeekNum; }
+#if PLATFORM_SHIELD
+    void setWeekNum(u32 weekNum) { mWeekNum = weekNum % 7; }
+#else
     void setWeekNum(int weekNum) { mWeekNum = weekNum; }
+#endif
```
TakaRikka, line 22: is this for sure real? id be surprised that other versions dont also do the modulo

include/d/actor/d_a_npc_cdn3.h
caseif: I agree it's weird, but it seems like the function signature actually did change in Shield and this was the only way I could manage to reconcile it.

src/JSystem/J3DGraphAnimator/J3DSkinDeform.cpp
caseif: I've revisited this function a few times over the months and so far I haven't been able to find any way of writing this logic that results in such drastic differences in the codegen, so I kind of suspect that it may have been refactored in the version of J3D used for the Shield version. That said, I agree it could still possibly be a fakematch and it was just written in some obscure way that we haven't figured out yet - I'll add a comment for good measure noting this.

include/d/actor/d_a_tag_schedule.h
caseif: As far as I can tell it's real - it seems to be a safeguard that was added some time after the GCN versions were built.

src/d/actor/d_a_obj_cblock.cpp
caseif: Looking again, this was copied as-is from the header and I have absolutely no idea where the `35.0f` value comes from since no other version has the HIO class present. That in mind I think it's best to remove it.

include/d/actor/d_a_npc_cdn3.h
LagoLunatic: What do you mean by "the function signature changed"? Between versions? It looks the same across versions to me.
image

Or do you mean that it changed between the symbol maps and the decomp? That could just mean the decomp itself is wrong. Ghidra likes to put float arguments in the wrong order, so someone could have copied the wrong argument order from Ghidra. Did you check if the argument order from ShieldD's maps still matches in all other versions?

caseif: My mistake, I assumed the order in the code was correct and figured it had changed without verifying against GCN/Wii (it actually doesn't even exist in GCN) - I think you're right about it being an issue with copy-pasting from Ghidra.

include/d/actor/d_a_tag_schedule.h
LagoLunatic: The safeguard still exists in GCN, but seems to be outside the inline.
image
Does it not match if you remove the ifdefs and put it inside the inline for all versions?

caseif: Good catch - there's an additional cast needed in the inline but after adding that and removing the extra modulos it seems to match for all versions.

PR #2841, opened 2025-11-21
npc_grm OK

maimai0254:
Most of daNpc_grM_c::evtProc() was copied from daNpcT_c::evtProc().
I have commented out the elements related to the debug build.

### DIFF
--- a/src/d/actor/d_a_npc_grm.cpp
+++ b/src/d/actor/d_a_npc_grm.cpp
@@ -428,7 +428,7 @@ BOOL daNpc_grM_c::evtProc() {
     dComIfG_play_c* play = &g_dComIfG_gameInfo.play;
     if (play->getEvent().runCheck()) {
 #else
-    if (dComIfGp_event_runCheck())
+    if (dComIfGp_event_runCheck()) {
 #endif
         if (eventInfo.checkCommandTalk()) {
             if (!checkChangeEvt()) {


src/d/actor/d_a_npc_grm.cpp
```diff
@@ -281,450 +70,236 @@ static char* l_resNameList[3] = {
...
+        // OS_REPORT("\t(%s:%d) flowNo:%d, MaxItem:%d, group:%d<%08x> ");
 
-/* 809D71B0-809D71BC 00031C 000C+00 1/1 0/0 0/0 .data            __vt__12dBgS_AcchCir */
-SECTION_DATA extern void* __vt__12dBgS_AcchCir[3] = {
-    (void*)NULL /* RTTI */,
-    (void*)NULL,
-    (void*)__dt__12dBgS_AcchCirFv,
-};
+        if (isDelete()) {
+            // OS_REPORT("===>isDelete:TRUE\n");
```
TakaRikka: these should be able to be uncommented?

src/d/actor/d_a_npc_grm.cpp
```diff
@@ -779,503 +421,456 @@ void daNpc_grM_c::srchActors() {
 }
 
 /* 809D4BBC-809D4E2C 000C3C 0270+00 1/0 0/0 0/0 .text            evtProc__11daNpc_grM_cFv */
-void daNpc_grM_c::evtProc() {
-    // NONMATCHING
+BOOL daNpc_grM_c::evtProc() {
+    BOOL ret = FALSE;
+// #if VERSION != VERSION_SHIELD_DEBUG
```
TakaRikka: all this should be able to be uncommented too?

maimai0254: I don't have the orig/ShieldD asset, so I'm unsure how to verify it against the debug code. In this case, should I remove the comment?

TakaRikka: in general yes if you cant confirm i'd just leave it out, though since a lot of that is copypasted in many npc's and it looks correct i'd just try compiling it anyway

maimai0254: Understood. I'll do just that.

PR #2842, opened 2025-11-21
d_a_b_mgn debug almost equivalent

carter-ktb21:


include/f_op/f_op_actor.h
```diff
@@ -332,13 +332,25 @@ class fopEn_enemy_c : public fopAc_ac_c {
     void onDownFlg() { mFlags |= fopEn_flag_Down; }
     void onHeadLockFlg() { mFlags |= fopEn_flag_HeadLock; }
 
+    #if DEBUG
+    void offWolfBiteDamage() { mFlags &= (u16)~fopEn_flag_WolfBiteDamage; }
+    #else
     void offWolfBiteDamage() { mFlags &= ~fopEn_flag_WolfBiteDamage; }
+    #endif
```
TakaRikka, line 339: does the cast not match for retail?

src/d/actor/d_a_b_mgn.cpp
```diff
@@ -273,7 +287,7 @@ int daB_MGN_c::JointCallBack(J3DJoint* i_joint, int param_1) {
 }
 
 /* 80610440-80610444 000008 0004+00 2/2 0/0 0/0 .bss             None */
-static u8 l_initHIO;
+static u8 l_initHIO_139_bss_0;
```
TakaRikka: why the name change?

include/f_op/f_op_actor.h
carco-games: The cast does not match

TakaRikka (approved):

PR #2849, opened 2025-11-23
d_a_npc_guard all functions OK 

Trueffeloot:
Can't be linked yet because of some weak function order problems


src/d/actor/d_a_npc_guard.cpp
```diff
@@ -1,1045 +1,665 @@
...
-    // NONMATCHING
-}
-
-/* 809F0148-809F0184 000468 003C+00 2/2 0/0 0/0 .text            __dt__4cXyzFv */
-// cXyz::~cXyz() {
-extern "C" void __dt__4cXyzFv() {
-    // NONMATCHING
+    if (m_path.isPath()) {
+        if (m_path.checkPoint(cXyz(current.pos), speedF)) {
+            if (m_path.checkPathEnd(cXyz(current.pos), speedF) && !m_path.isClose()) {
```
caseif: Explicit `cXyz` constructor isn't necessary here, also according to debug `m_path.isClose()` should be in a separate `if` statement.

Trueffeloot:
Applied what you mentioned

PR #2850, opened 2025-11-23
Fix more nonmatchings

LagoLunatic:


include/dol2asm.h
```diff
@@ -78,6 +78,7 @@
 #define INFINITY ((float)(_HUGE_ENUF * _HUGE_ENUF))
 #define HUGE_VAL ((double)INFINITY)
 #define HUGE_VALL ((long double)INFINITY)
+static const float INF = 2000000000.0f;
```
TakaRikka: maybe move this to global.h? i plan to nuke dol2asm.h as soon as possible

PR #2852, opened 2025-11-24
d_a_alink_HIO work

Pheenoh:


src/d/actor/d_a_alink.cpp
```diff
@@ -2960,7 +2955,11 @@ s16 daAlink_c::getNeckAimAngle(cXyz* param_0, s16* param_1, s16* param_2, s16* p
     s16 sp18;
     s16 sp16 = field_0x2fe6 + mBodyAngle.y;
     if ((mProcID == PROC_GOAT_CATCH && mProcVar1.field_0x300a == 0) || (mProcID == PROC_HAND_PAT && mProcVar2.field_0x300c == 0)) {
+#ifdef DEBUG
```
TakaRikka: might want to use `VERSION == VERSION_SHIELD_DEBUG` instead since this doesnt seem like a debug specific thing

PR #2858, opened 2025-11-24
d_a_grass Matching

hatal175:


tools/converters/matDL_dis.py
```diff
@@ -330,7 +330,7 @@ def convert_binary_to_matDL_c_source(src_path, dest_path, symbol_name, scope):
     with open(src_path, "rb") as binary_file, open(dest_path, "w") as c_file:
         c_file.write("#ifndef LOAD_BP_REG\n")
         c_file.write("#define U32_AS_U8(v) ((u8)((v) >> 24)), ((u8)((v) >> 16)), ((u8)((v) >> 8)), ((u8)((v) >> 0))\n")
-        c_file.write("#define U24_AS_U8(v) ((u8)((v) >> 16)), ((u8)((v) >> 8)), ((u8)((v) >> 0))\n")
+        c_file.write("#define U24_AS_U8(v) (((v) >> 16) & 0xff), (((v) >> 8)& 0xff), (((v) >> 0) & 0xff)\n")
```
LagoLunatic, line 333: So this is what affects sinit regalloc?
Maybe the other two lines should also be changed to use &0xFF then.

hatal175: Maybe... I don't think they're unmatched anywhere else so it doesn't particularly matter

PR #2863, opened 2025-11-25
d_a_npc debug work

caseif:
This PR mostly matches `d_a_npc`'s functions and fixes up the file structure to more closely match what was likely the original structure based on function ordering and debug info. `d_a_npc3.cpp` appears not to be real and is now removed (its contents are now in `d_a_npc2.cpp` instead), and functions implemented in `d_a_npc2.cpp` and `d_a_npc4.cpp` have been moved to their own respective headers.


hatal175:
note conflict

hatal175 (approved):

PR #2864, opened 2025-11-25
d_a_e_mk_bo ShieldD work

ItsNiklas:
- Big Derust
- Some `d_a_e_mk_bo` debug matching
- text section matching now
- ShieldD +0.05%

### DIFF
--- a/src/d/actor/d_a_e_mk_bo.cpp
+++ b/src/d/actor/d_a_e_mk_bo.cpp
@@ -186,14 +186,14 @@ static s8 e_mk_bo_shot(e_mk_bo_class* i_this) {
 /* 8071D694-8071DBA0 000A94 050C+00 1/1 0/0 0/0 .text            e_mk_bo_start__FP13e_mk_bo_class */
 static s8 e_mk_bo_start(e_mk_bo_class* i_this) {
     fopAc_ac_c* actor = &i_this->enemy;
-    e_mk_class* e_mk_ = (e_mk_class*)fopAcM_SearchByID(actor->parentActorID);
+    fopAc_ac_c* a_parent = fopAcM_SearchByID(actor->parentActorID);
 
     s8 rt = 2;
-    if (e_mk_ == NULL) {
+    if (a_parent == NULL) {
         return rt;
     }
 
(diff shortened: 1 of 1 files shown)


src/d/actor/d_a_e_mk_bo.cpp
```diff
@@ -184,16 +186,18 @@ static s8 e_mk_bo_shot(e_mk_bo_class* i_this) {
 /* 8071D694-8071DBA0 000A94 050C+00 1/1 0/0 0/0 .text            e_mk_bo_start__FP13e_mk_bo_class */
 static s8 e_mk_bo_start(e_mk_bo_class* i_this) {
     fopAc_ac_c* actor = &i_this->enemy;
-    e_mk_class* e_mk = (e_mk_class*)fopAcM_SearchByID(actor->parentActorID);
+    e_mk_class* e_mk_ = (e_mk_class*)fopAcM_SearchByID(actor->parentActorID);
```
hatal175: I think that all the fopAcM_SearchByID results are just fopAc_ac_c*. Possibly still with (fopAc_ac_c*). Any chance you can try that? It seems more likely than the current way.

src/d/actor/d_a_e_mk_bo.cpp
```diff
@@ -717,11 +726,15 @@ static int daE_MK_BO_IsDelete(e_mk_bo_class* i_this) {
     return 1;
 }
 
+static void dummy() {
```
hatal175: This is not a dummy function - it's the use of the fopAcM_RegisterDeleteID macro in daE_MK_BO_Delete. Using the complementary fopAcM_RegisterCreateID macro in daE_MK_BO_Create is also likely.

ItsNiklas: I see, thank you. There is no "Create -> " strings in debug `.data` so I think it is not used?

src/d/actor/d_a_e_mk_bo.cpp
ItsNiklas: Good call, this makes way more sense. And yes cast is required for some GCN regswaps.

TakaRikka (approved):

PR #2866, opened 2025-11-25
d_a_npc_rafrel debug almost matching

carter-ktb21:
was hoping that somehow matching more of debug would shed light on the weak func issue, but no luck


TakaRikka:
conflict needs to be fixed

PR #2871, opened 2025-11-25
d_a_obj_carry ALL_GCN matching

ItsNiklas:
- d_a_obj_carry now matching (_+0.02%_)
- compiler generates a temporary stack slot for the cast
- yeah idk


src/d/actor/d_a_obj_carry.cpp
```diff
@@ -1913,7 +1910,7 @@ int daObjCarry_c::execute() {
 
     field_0xda9 = fopAcM_checkCarryNow(this);
     field_0xdac = mCyl.ChkCoHit();
-    field_0xdad = mAcch.ChkWaterIn();
+    field_0xdad = (bool) mAcch.ChkWaterIn();
```
TakaRikka: does this match with something like a ternary or `!= 0` instead of a cast?

ItsNiklas: This was only to get closer to debug, but I solved it now.

hatal175 (approved):

PR #2872, opened 2025-11-26
d_kankyo matching

caseif:
`d_kankyo` is matching for all GCN versions with this PR. There are also some changes to fix RTTI ordering in debug - most notably I moved `dKy_tevStr_c` to its own header because debug indicates that `f_op_actor.h` is likely included before `d_kankyo.h`, but `fopAc_ac_c` contains a `dKy_tevStr_c` member which would otherwise preclude this order.


include/d/dolzel_base.pch
```diff
@@ -14,10 +14,11 @@
 #include "string.h" // IWYU pragma: export
 #include "JSystem/JUtility/JUTGamePad.h" // IWYU pragma: export
 #include "JSystem/J3DGraphAnimator/J3DMaterialAnm.h" // IWYU pragma: export
+#include "JSystem/JHostIO/JORReflexible.h" // IWYU pragma: export
 #include "SSystem/SComponent/c_m3d_g_pla.h" // IWYU pragma: export
 #include "SSystem/SComponent/c_bg_s_chk.h" // IWYU pragma: export
 #include "f_op/f_op_actor.h" // IWYU pragma: export
-#include "d/d_bg_s_poly_pass_chk.h" // IWYU pragma: export
+#include "d/d_bg_s_gnd_chk.h"
```
hatal175: Add the IWYU comment

PR #2880, opened 2025-11-28
d_a_obj_crvgate debug almost matching

ItsNiklas:
Only string pooling/RTTI issues remain.
Surprisingly, I think this class uses `S16Vec`, since the obvious xyz accesses but no `csXyz` in the debug class constructor.

### DIFF
--- a/src/d/actor/d_a_obj_crvgate.cpp
+++ b/src/d/actor/d_a_obj_crvgate.cpp
@@ -3,8 +3,8 @@
  *
  */
 
- #include "d/dolzel_rel.h"  // IWYU pragma: keep
- 
+#include "d/dolzel_rel.h"  // IWYU pragma: keep
+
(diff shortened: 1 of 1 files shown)


TakaRikka:
conflict needs to be fixed

src/d/actor/d_a_obj_crvgate.cpp
```diff
@@ -99,13 +98,17 @@ int daObjCRVGATE_c::checkOpen() {
     }
 
     daPy_py_c* player = daPy_getPlayerActorClass();
-    s16 height_diff = fopAcM_GetAngle_p(this)->y - fopAcM_GetAngle_p(player)->y;
+    s16 height_diff = this->current.angle.y - player->current.angle.y;
```
TakaRikka: can remove the `this->`

TakaRikka (approved):

PR #2886, opened 2025-11-28
Adding tons of enum resource files

roeming:
I was able to iterate over all the currently supported versions and extract enums for all of the arc files, exported to the `./assets/{VERSION}/res/` folder. the `res_arc.py` tool can now take input and output files, but is not currently hooked into the build steps. `configure.py` was updated to add the `assets/{VERSION}/` path for includes, so version specific assets can be included.


roeming:
There are lots of enums that are shared between all versions, of the roughly 10,000 files for enums, there are only 1,500 unique files. That includes files that are duplicated across arcs. There's probably a nicer way to handle it, but with auto-extraction, I can't think of a nice way to extract to the correct folder (shared or unique). Just something I stumbled across

TakaRikka (approved):

PR #2895, opened 2025-11-30
project cleanup

TakaRikka:
- removes dol2asm.h and all old unneeded dol2asm comments and macros
- removes function address comments to make things less GC USA specific
- fixes macros like attribute align and declspec weak for compiler specific syntax
- adds more mwcc ifdef wraps around mwcc specific things


Pheenoh (commented):
LGTM outside of broken functions/regressions if you know why those happened?

PR #2899, opened 2025-12-01
d_a_e_dn work

carter-ktb21:


src/d/actor/d_a_e_dn.cpp
```diff
@@ -14,14 +14,18 @@
 #include "f_op/f_op_actor_enemy.h"
 #include "Z2AudioLib/Z2Instances.h"
 
+#if DEBUG
+#pragma nosyminline on
```
TakaRikka: does debug in general benefit from nosyminline? it might be better to have this as a general debug compile flag rather than manually defining it in the TU

hatal175:
note conflict

TakaRikka (approved):

PR #2900, opened 2025-12-01
Some variable names for hookshot and moved some values

oga-sven:
First PR. just a small contribution towards hookshot variable names and putting values in the data.inc. Im new to CPP so feedback is welcome!

### DIFF
--- a/include/d/actor/d_a_alink.h
+++ b/include/d/actor/d_a_alink.h
@@ -1973,9 +1973,7 @@ public:
     /* 0x64 */ f32 mBossShootSpeed;
     /* 0x68 */ f32 mBossReturnSpeed;
     /* 0x6C */ f32 mBossStickReturnSpeed;
-    /* 0x70 */ f32 mHookTipAnimationSteps;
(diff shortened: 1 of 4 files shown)


include/d/actor/d_a_alink.h
```diff
@@ -1972,7 +1973,9 @@ class daAlinkHIO_hookshot_c1 {
     /* 0x64 */ f32 mBossShootSpeed;
     /* 0x68 */ f32 mBossReturnSpeed;
     /* 0x6C */ f32 mBossStickReturnSpeed;
-};  // Size: 0x70
+    /* 0x70 */ f32 mHookTipAnimationSteps;
+    /* 0x74 */ f32 mLockCursorSpinnerSpeed;
+};  // Size: 0x78
```
TakaRikka: this data shouldnt be added here, it needs to stay identical to the original to keep matching the original binary. im assuming this is what's breaking all those matches

oga-sven: ah got it! ill correct it. mustve misunderstood that one :)

PR #2904, opened 2025-12-02
d_a_obj_damcps debug equivalent

ItsNiklas:
- Create `d_a_obj_damcps`.
- Debug-only implementation.
- I had to fully create it from scratch so idk if there are mistakes.
- Might need mDoMtx comments.
- .text matches 100%, .rodata normal debug lbl missing.
- Func order does not match, `daObjDamCps_c::daObjDamCps_c()` and `daObjDamCps_HIO_c::~daObjDamCps_HIO_c()` are wrongly placed.
- `actor_process_profile_definition` ifdefs are not nice, but I didn't find a nicer option. I could completly copy it over but only the sub_method needs to be changed. I wonder if there is a better solution.

### DIFF
--- a/src/d/actor/d_a_obj_damCps.cpp
+++ b/src/d/actor/d_a_obj_damCps.cpp
@@ -40,7 +40,7 @@ void daObjDamCps_HIO_c::genMessage(JORMContext* ctx) {
     // Damage Cylinder
     ctx->genLabel("ダメージ円柱", 0, 0, NULL, 0xffff, 0xffff, 0x200, 0x18);
     // Adjustment
-    ctx->genCheckBox("調整", &mAdj, '\x01', 0, NULL, 0xffff, 0xffff, 0x200, 0x18);
+    ctx->genCheckBox("調整", &mAdj, 1, 0, NULL, 0xffff, 0xffff, 0x200, 0x18);
     // X Angle
     ctx->genSlider("Ｘ角度", &mAngle.x, -0x7fff, 0x7fff, 0, NULL, 0xffff, 0xffff, 0x200, 0x18);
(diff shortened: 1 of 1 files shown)


src/d/actor/d_a_obj_damCps.cpp
```diff
@@ -1,25 +1,221 @@
...
+    mAdj = false;
+    mAngle.setall(0x0);
+    mScale.setall(1.0f);
+}
+
+void daObjDamCps_HIO_c::genMessage(JORMContext* ctx) {
+    // Damage Cylinder
+    ctx->genLabel("ダメージ円柱", 0, 0, NULL, 0xffff, 0xffff, 0x200, 0x18);
+    // Adjustment
+    ctx->genCheckBox("調整", &mAdj, '\x01', 0, NULL, 0xffff, 0xffff, 0x200, 0x18);
```
hatal175: Just 1 or TRUE I think

src/d/actor/d_a_obj_damCps.cpp
```diff
@@ -1,25 +1,221 @@
...
+    fopAcM_setCullSizeBox(this, l_cull_box.min.x * scale.x, l_cull_box.min.y * scale.y,
+                          l_cull_box.min.z * scale.x, l_cull_box.max.x * scale.x,
+                          l_cull_box.max.y * scale.y, l_cull_box.max.z * scale.x);
+    return 1;
+}
+
+int daObjDamCps_c::create() {
+    fopAcM_ct(this, daObjDamCps_c);
+
+    if (Create() == 0) {
```
hatal175: Does !Create() work?

PR #2909, opened 2025-12-03
Match 4 more functions

### LINKED ISSUE #678: d_a_npc_ykw.cpp

### LINKED ISSUE #739: d_a_npc_ykm.cpp

LagoLunatic:
closes #739
closes #678

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -805,7 +805,7 @@ config.libs = [
             Object(MatchingFor(ALL_GCN), "d/d_meter2.cpp"),
             Object(MatchingFor(ALL_GCN), "d/d_msg_out_font.cpp"),
             Object(MatchingFor(ALL_GCN), "d/d_msg_class.cpp"),
-            Object(Equivalent, "d/d_msg_object.cpp"),  # weak func order
+            Object(Equivalent, "d/d_msg_object.cpp"), # weak func order
             Object(MatchingFor("GZ2P01", "GZ2J01"), "d/d_msg_unit.cpp"),
             Object(MatchingFor(ALL_GCN), "d/d_msg_scrn_3select.cpp"),
             Object(MatchingFor(ALL_GCN), "d/d_msg_scrn_arrow.cpp"),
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -2047,7 +2047,7 @@ def MatchingFor(*versions) -> bool:
     ActorRel(MatchingFor(ALL_GCN), "d_a_npc_soldierA"),
     ActorRel(MatchingFor(ALL_GCN), "d_a_npc_soldierB"),
     ActorRel(MatchingFor(ALL_GCN), "d_a_npc_sq"),
-    ActorRel(NonMatching, "d_a_npc_the"),
+    ActorRel(Equivalent, "d_a_npc_the"), # weak func order (sinShort)
```
YunataSavior, line 2050: If this is the same weak func order issue as the rest of the NPCF non-matching TUs, then we should annotate all of them identically.

LagoLunatic: I already commented `weak func order (sinShort)` on all the ones I know about in a previous PR. Are there more of them missing the comment? I haven't looked at every NPC.

YunataSavior: Yes:
* d_a_npc_impal
* d_a_npc_ins
* d_a_npc_rafrel
* d_a_npc_shad

Semi-related, d_a_npc_tks can be marked as `Equivalent`. It too is stuck in sinShort weak func order purgatory.

YunataSavior:
BTW, my VSCode complains if there's only 1 space between the end of the functional line and the comment. That's why I started putting 2 spaces.

PEP 8 guideline:
https://peps.python.org/pep-0008/

configure.py
LagoLunatic: Seems `d_a_obj_tks` also has the same issue. Interesting that they all inherit from `daNpcF_c` since that's also the class that had the ctor that we had to force not to inline. And sinShort is also supposed to be inline but doesn't getting inlined in these TUs. I'd guess that the same root cause behind the ctor issue is behind all the sinShort issues.

LagoLunatic:
> BTW, my VSCode complains if there's only 1 space between the end of the functional line and the comment. That's why I started putting 2 spaces.

flake8 complains about dozens of different things in configure.py for no reason, so just using two spaces is not enough to get rid of the errors. You can just disable the addon if it bothers you,

PR #2911, opened 2025-12-03
Resolve all "illegal jump past declaration/initializer" warnings

ItsNiklas:
- Fix all current `illegal jump past initializer` MWCC warnings (clang `switch_into_protected_scope` errors) around switches. Exact method depends on debug, new local scope, moving variable declarations, or a separate initializer.
- Now compiles with hundreds of warnings less.
- Almost always this caused debug improvements. 
- Lots of minor debug work when I felt like it.

### DIFF
--- a/src/JSystem/JUtility/JUTCacheFont.cpp
+++ b/src/JSystem/JUtility/JUTCacheFont.cpp
@@ -277,7 +277,7 @@ void JUTCacheFont::setBlock() {
             break;
         case 'GLY1':
             memcpy(piVar5, pData, 0x20);
-            JKRAramBlock* iVar1; // Force end-of-scope branch
+            JKRAramBlock* iVar1;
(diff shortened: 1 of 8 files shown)


src/JSystem/JUtility/JUTCacheFont.cpp
```diff
@@ -277,9 +277,9 @@ void JUTCacheFont::setBlock() {
             break;
         case 'GLY1':
             memcpy(piVar5, pData, 0x20);
-            JKRAramBlock* iVar1 =
-                JKRMainRamToAram((u8*)pData + 0x20, aramAddress, pData[1] - 0x20,
-                                 EXPAND_SWITCH_UNKNOWN0, 0, NULL, 0xffffffff, NULL);
+            JKRAramBlock* iVar1; // Force end-of-scope branch
```
TakaRikka: i think its better to leave out the comments for these imo unless it seems like an obvious fakematch

PR #2915, opened 2025-12-05
Various debug conditional compilation cleanup

caseif:
- Define `DEBUG` as `0` if it's not already defined (via `global.h`)
- Updates all occurrences of `#ifdef`/`ifndef DEBUG` accordingly
- Clean up condition expressions which were previously awkward due to guards being placed in the middle of them - the `DEBUG` macro can now be used inline instead
- Clean up some unnecessary guards around `JUT_ASSERT`, `OS_REPORT`, etc.
- Replace guarded `OSReport` with `OS_REPORT` where possible
- Remove unnecessary guards around debug functions which are ultimately stripped anyway

### DIFF
--- a/src/d/actor/d_a_obj_kgate.cpp
+++ b/src/d/actor/d_a_obj_kgate.cpp
@@ -259,9 +259,7 @@ int daObjKGate_c::Create() {
                           l_cull_box.max.y, l_cull_box.max.z);
 
     if (dComIfG_Bgsp().Regist(mpBgW, this)) {
-#if DEBUG
-        OSReport_Error("キコルの門：ＢＧ登録失敗しました\n");
-#endif
+        OS_REPORT_ERROR("キコルの門：ＢＧ登録失敗しました\n");
         return 0;
     }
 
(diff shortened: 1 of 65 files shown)


src/d/actor/d_a_obj_kgate.cpp
```diff
@@ -336,7 +336,7 @@ int daObjKGate_c::create1st() {
         phase_state =
             MoveBGCreate(l_arcName[mNameArg], l_gateDzbIdx[mNameArg], NULL, heap_size, NULL);
         if (phase_state == cPhs_ERROR_e) {
-#ifdef DEBUG
+#if DEBUG
             OSReport("KGATE PARAM:%x\n", fopAcM_GetParam(this));
 #endif
```
ItsNiklas, line 339: There are a bunch of `OSReport` lines remaining, are these not possible to replace like you wrote?

roeming: You could replace those calls with the already existing `OS_REPORT`, which only calls `OSReport` in debug

caseif: Thanks for pointing that out, I must have missed those occurrences originally!

caseif:
This may conflict with #2917 - that should probably be merged first and then I can rebase.

TakaRikka (approved):

PR #2916, opened 2025-12-05
d_a_e_gb debug + documentation

carter-ktb21:


src/d/actor/d_a_e_gb.cpp
```diff
@@ -106,398 +135,408 @@ static u8 data_806C7928;
...
-    s8 sVar1 = 0;
-
-    switch (i_this->field_0x69c) {
+    cXyz work, target_offset;
+    f32 offset = TREG_F(12) + 30.0f;
+    #if VERSION == VERSION_SHIELD_DEBUG
+    f32 pl_dist = (actor->home.pos - player->current.pos).abs();
+    #else
+    f32 pl_dist = (actor->home.pos - dComIfGp_getPlayer(0)->current.pos).abs();
+    #endif
```
TakaRikka, line 146: these seem like a fake match

src/d/actor/d_a_e_gb.cpp
```diff
@@ -106,398 +135,408 @@ static u8 data_806C7928;
...
+            MtxPosition(&work, &i_this->currentPosTarget);
+            i_this->currentPosTarget += actor->home.pos;
+
+            cLib_addCalc2(&actor->current.pos.x, i_this->currentPosTarget.x, 0.1f, NREG_F(3) + 80.0f);
+            cLib_addCalc2(&actor->current.pos.y, i_this->currentPosTarget.y, 0.1f, NREG_F(3) + 80.0f);
+            cLib_addCalc2(&actor->current.pos.z, i_this->currentPosTarget.z, 0.1f, NREG_F(3) + 80.0f);
+
+            cLib_addCalcAngleS2(&actor->current.angle.y, -0x4000 /* -90° */, 16, 0x200 /* ~2.8° */);
+            cLib_addCalcAngleS2(&actor->shape_angle.y, 0x4000 /* 90° */, 16, 0x200 /* ~2.8° */);
+            cLib_addCalcAngleS2(&actor->shape_angle.x, 0x2000 /* 45° */, 16, 0x100 /* ~1.4° */);
```
TakaRikka: im not totally sure about the degree comments, maybe we could use a macro but want to think about it some more first. also since the last parameters are steps i dont really like using degrees for them

src/d/actor/d_a_e_gb.cpp
carco-games: It's happened before in another TU, but the asm genuinely seems to indicate different positions of the getPlayer call. Not sure how to overcome that without the #if

src/d/actor/d_a_e_gb.cpp
carco-games: I get that. I personally think we should come up with something, as thinking of the values in degrees really helps visualize what's happening in regard to angles and rotations.

Also, the last parameters as degrees is due to the fact that each step is a slight rotation in degrees to the target angle.

TakaRikka (approved):

PR #2920, opened 2025-12-06
Improving compatibility with standard cpp compiler

kipcode66:
The `NDEBUG` not being defined while being used inside of an `if` statement causes some problems, So I added an additional define which is always defined for use in the `if` statement, while not interfering with the current usage of the `DEBUG` and `NDEBUG` macros.

### DIFF
--- a/src/d/d_ev_camera.cpp
+++ b/src/d/d_ev_camera.cpp
@@ -995,15 +995,9 @@ bool dCamera_c::transEvCamera(int param_1) {
             trans->mEye = (*ptr1).mEye;
             trans->mFovy = (*ptr1).mFovy;
             trans->field_0x3c = (*ptr1).mBank;
-#ifdef __MWERKS__
-            if ((*ptr1).mBank != 0.0f) {
+            if (float(s16(ptr1->mBank)) != 0.0f) {
                 trans->field_0x68 = true;
             }
-#else
-            if ((float)ptr1->mBank.Val() != 0.0f) {
-                trans->field_0x68 = true;
-            }
-#endif
         }
 
         if (!getEvIntData(&slot, "StartSlot")) {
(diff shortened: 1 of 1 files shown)


TakaRikka:
should be good once d_stage conflict is fixed

src/d/actor/d_a_alink_hang.inc
```diff
@@ -2533,9 +2533,9 @@ void daAlink_c::setRoofHangHandOnSE(cBgS_PolyInfo* unused) {
 int daAlink_c::checkRoofHangMovePos() {
     static const s16 offsetAngle[] = {
         0x0000,
-        0x8000,
+        -0x8000,
         0x4000,
-        0xC000,
+        -0xC000,
```
magcius: I think this should be -0x4000?

src/d/d_ev_camera.cpp
```diff
@@ -995,7 +995,7 @@ bool dCamera_c::transEvCamera(int param_1) {
             trans->mEye = (*ptr1).mEye;
             trans->mFovy = (*ptr1).mFovy;
             trans->field_0x3c = (*ptr1).mBank;
-            if ((*ptr1).mBank != 0.0f) {
+            if ((float)ptr1->mBank.Val() != 0.0f) {
```
TakaRikka: are these necessary for complying with the standard? debug doesnt show any inlines being used, just that mBank uses its s16 operator and then its converted into a float. if that doesn't comply, then i'd use the `__MWERKS__` define there and then add the changes as an else so that it doesnt break any version matching

kipcode66: It is indeed needed for compiling with standard, if we use mBank directly, The compiler complains that it can't find an implementation of the operator for cSAngle and float.
```
/Users/kipcode66/git/dusk/src/d/d_ev_camera.cpp:998:31: error: use of overloaded operator '!=' is ambiguous (with operand types 'cSAngle' and 'float')
  998 |             if ((*ptr1).mBank != 0.0f) {
      |                 ~~~~~~~~~~~~~ ^  ~~~~
```
So to do the expected operation of comparing the s16 value of angle with the float 0.0f, I need to extract it somehow.

I'll use the `__MWERKS__` macro to make the change in that case.

magcius: `float(s16(ptr1->mBank)) != 0.0f` should work then, no?

kipcode66: I forgot the operator for converting to s16 was implemented in the class. I'll do that, thanks!

PR #2925, opened 2025-12-07
match debug for D_menu_fmap2_d

roeming:
the only one that caused some trouble was `checkPlayerWarpAccept`, which looks really similar but I couldn't match between debug and release, and may return a `u8` but caused linking issues elsewhere. The other major change was `dSv_event_flag_c::saveBitLabels` needed to be `s16` to match debug, so I locked it behind a define for `PLATFORM_SHIELD`, although it looks like Wii can match with either `s16` or `u16`, but GC needs `u16`.

### DIFF
--- a/include/d/actor/d_a_npc_cdn3.h
+++ b/include/d/actor/d_a_npc_cdn3.h
@@ -223,7 +223,7 @@ public:
                 break;
             }
                 /* dSv_event_flag_c::F_0281 - Shop - Malo Mart opens in Castle Town */
-            if (dComIfGs_isEventBit((u16)dSv_event_flag_c::saveBitLabels[281]) && uVar3 == 9) {
+            if (dComIfGs_isEventBit(dSv_event_flag_c::saveBitLabels[281]) && uVar3 == 9) {
                 uVar3 = 0xb;
             }
         } else {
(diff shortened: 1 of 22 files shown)


src/d/d_menu_fmap2D.cpp
```diff
@@ -1907,16 +2014,18 @@ void dMenu_Fmap2DBack_c::drawDebugStageArea() {
...
+                    J2DFillBox(v - 3.0f, v2 - 3.0f, 6.0f, 6.0f, colors[stage_no % 6]);
                 }
             }
         }
     }
 }
 
 void dMenu_Fmap2DBack_c::drawDebugRegionArea() {
-    getAllMapZoomRate();
+    if (this->mZoom != getAllMapZoomRate()) {
```
TakaRikka: `this->` can be removed?

src/d/d_menu_fmap2D.cpp
```diff
@@ -1995,27 +2104,30 @@ void dMenu_Fmap2DBack_c::calcDrawPriority() {
...
             }
-            field_0x1230[i] = dVar2;
-            dVar2 = region;
         }
     }
 
     if (mRegionCursor != 6) {
         for (int i = 7; i >= 0; i--) {
-            region = field_0x1230[i];
+            // region = field_0x1230[i];
```
TakaRikka: unneeded comment?

PR #2939, opened 2025-12-10
Resource

kipcode66:
Fix templates and resource.cpp

### DIFF
--- a/include/JSystem/JGadget/define.h
+++ b/include/JSystem/JGadget/define.h
@@ -15,14 +15,14 @@ public:
     JGadget_outMessage(MessageFunc fn, const char* file, int line);
     ~JGadget_outMessage();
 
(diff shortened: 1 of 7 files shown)


include/JSystem/JGadget/define.h
```diff
@@ -15,14 +15,14 @@ class JGadget_outMessage {
     JGadget_outMessage(MessageFunc fn, const char* file, int line);
     ~JGadget_outMessage();
 
-    JGadget_outMessage& operator<<(int param_1) { return *this << (s32)param_1; }
+    JGadget_outMessage& operator<<(int param_1) { return *this << (signed long)param_1; }
```
roeming: Why the change? `s32` is typedef'd to `signed long`.

PR #2944, opened 2025-12-11
d_particle debug work

carter-ktb21:


include/d/d_com_inf_game.h
```diff
@@ -838,8 +839,57 @@ class dComIfG_inf_c {
     dComIfG_inf_c() { this->ct(); }
     ~dComIfG_inf_c() {}
     void ct();
+    void createBaseCsr();
     dComIfG_play_c& getPlay() { return play; }
 
+#if DEBUG
```
TakaRikka: all of this Csr and listCursor stuff should be `#if PLATFORM_WII || VERSION == VERSION_SHIELD_DEBUG`, its not an actual debug feature

carco-games: I had a hunch that might be the case. I'll change it!

PR #2946, opened 2025-12-11
Link J2DWindowEX

roeming:
I was able to link this because of a non-const copy ctor for `J2DGXColorS10`. This doesn't appear in debug, but fixed all issues in GC and didn't break anywhere else. @LagoLunatic found in TWW maps that the `J3DGXColorS10` had a non-const copy ctor, this was likely a bug that permeated the GC versions, but was fixed in later versions, hence why we couldn't find it in debug.

### DIFF
--- a/include/JSystem/J2DGraph/J2DMatBlock.h
+++ b/include/JSystem/J2DGraph/J2DMatBlock.h
@@ -27,6 +27,13 @@ struct J2DGXColorS10 : public GXColorS10 {
         b = other.b;
         a = other.a;
     }
+    
+    J2DGXColorS10(GXColorS10& other) {
+        r = other.r;
+        g = other.g;
+        b = other.b;
+        a = other.a;
+    }
 #else
     J2DGXColorS10(const J2DGXColorS10& other) {
         r = other.r;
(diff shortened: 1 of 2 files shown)


LagoLunatic:
In addition to:
```
J2DGXColorS10(J2DGXColorS10&)
J3DGXColorS10(J3DGXColorS10&)
```
These ones should also have no const for GCN I believe:
```
J2DGXColorS10(GXColorS10&)
J3DGXColorS10(GXColorS10&)
J3DGXColor(J3DGXColor&)
J3DGXColor(GXColor&)
```
Most of these are known because of TWW demo debug maps. A couple aren't used in TWW so it's not 100% sure, but as far as I can tell the version difference seems to apply to all JSystem color copy ctors.

roeming:
> In addition to:
> 
> ```
> J2DGXColorS10(J2DGXColorS10&)
> J3DGXColorS10(J3DGXColorS10&)
> ```
> 
> These ones should also have no const for GCN I believe:
> 
> ```
> J2DGXColorS10(GXColorS10&)
> J3DGXColorS10(GXColorS10&)
> J3DGXColor(J3DGXColor&)
> J3DGXColor(GXColor&)
> ```
> 
> Most of these are known because of TWW demo debug maps. A couple aren't used in TWW so it's not 100% sure, but as far as I can tell the version difference seems to apply to all JSystem color copy ctors.

I'm not sure about the `J3DGXColor` ones, they seem to break other pieces. It could be the case that we fakematched them, but I might just leave a note and skip those for now.

TakaRikka (approved):

PR #2947, opened 2025-12-12
d_a_obj_bemos debug work

carter-ktb21:


src/d/actor/d_a_obj_bemos.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include "SSystem/SComponent/c_lib.h"
 #include "d/actor/d_a_obj_bemos.h"
+#include "../assets/GZ2E01/res/Object/Obj_bm.h"
```
LagoLunatic: This is incorrect, it isn't supposed to hardcode the version. The include is supposed to be a different file for each version of the game due to version differences.
```suggestion
#include "res/Object/Obj_bm.h"
```

src/d/actor/d_a_obj_bemos.cpp
```diff
@@ -38,98 +43,189 @@ s16 const daObjBm_c::M_dir_base[4] = {
...
+#define AFTER_SWITCH_ACTIVATE_NO_SEARCH_TIME l_HIO.after_switch_activate_no_search_time
+#define DETECTION_ANGLE l_HIO.detection_angle
+#define SEARCH_SPD_ON_DETECT l_HIO.search_spd_on_detect
+#define BODY_SCALE l_HIO.body_scale
+#define SOUND_POS_MOVE_SPD l_HIO.sound_pos_move_spd
+#define NO_SEARCH_ATTACK_TIME l_HIO.no_search_attack_time
+#define DEBUG_DRAW l_HIO.debug_draw
+#define CHECK_DRAW l_HIO.check_draw
+#else
+#define BEAM_SCALE_X 0.5f
```
YunataSavior, line 210: These defaults are also the defaults of the l_HIO member fields. I'd prefer we extracted these constants into other defines, e.g. `DEFAULT_BEAM_SCALE_X`, then use those defaults in the dbg constructor + here.

carco-games: This would not work, as the macros are used in different ways. The retail simply calls the value, whereas the debug calls l_HIO.###

If I defined a macro to use values in the dbg constructor, I'd have to define separate macros for l_HIO calls, and it would just be messier

PR #2954, opened 2025-12-14
Npc

kipcode66:
match d_a_npc_cd2
Fix missing short conversion in d_a_movie_player (this will reduce slightly the % for gcn, but it will be functionally equivalent the actual code, contrarily to what the current code does)

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -756,7 +756,7 @@ config.libs = [
             Object(MatchingFor(ALL_GCN), "d/actor/d_a_no_chg_room.cpp"),
             Object(NonMatching, "d/actor/d_a_npc.cpp"),
             Object(MatchingFor(ALL_GCN), "d/actor/d_a_npc_cd.cpp"),
-            Object(MatchingFor(ALL_GCN), "d/actor/d_a_npc_cd2.cpp"), # stripped vtable order
+            Object(NonMatching, "d/actor/d_a_npc_cd2.cpp"), # stripped vtable order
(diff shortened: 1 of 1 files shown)


LagoLunatic:
IIRC, the reason `daNpcCd2_HIO_c`'s ctor was separated from the data that should go inside of it is because if you move the ctor up like you did it messes up function order, and if you move the data down it messes up data order. The bot doesn't report changes in function/data order, did you find a way to fix this, or did the order regress with this PR? You can run this command locally to check the function order (doesn't work well for data order so check that manually):
```
./tools/utilities/weak_order_diff.py d/actor/d_a_npc_cd2.o
```

Also, if you're moving cd2's data to be in-function statics anyway, there are two more of them: `a_transScaleTbl$3998` and `a_jntNumTbl$4692`. I think those two can be fixed without messing up the order unlike the other four.

src/d/actor/d_a_movie_player.cpp
```diff
@@ -3259,7 +3259,7 @@ static void daMP_MixAudio(s16* destination, s16*, u32 sample) {
 					r_mix = 32767;
 
                 if (JASDriver::getOutputMode() == 0) {
-                    l_mix = r_mix = ((r_mix >> 1) + (l_mix >> 1));
+                    l_mix = r_mix = (s16)((r_mix >> 1) + (l_mix >> 1));
```
TakaRikka, line 3262: i'm mixed on this change since it doesnt actually fully match debug either. feels like theres more that we're missing to fully solve it

PR #2961, opened 2025-12-16
Match some debug objects

ItsNiklas:


src/d/actor/d_a_obj_usaku.cpp
```diff
@@ -11,16 +11,18 @@
 #include "d/d_procname.h"
 
 static int daObj_Usaku_Draw(obj_usaku_class* i_this) {
-    g_env_light.settingTevStruct(0x10, &i_this->current.pos, &i_this->tevStr);
-    g_env_light.setLightTevColorType_MAJI(i_this->mpModel, &i_this->tevStr);
+    fopAc_ac_c* a_this = static_cast<fopAc_ac_c*>(i_this);
```
Luzagar: it should be fopAc_ac_c* actor = &i_this->actor; don't use static cast unless it's necessary to match

PR #2963, opened 2025-12-16
fix ENABLE_REGHIO macro usage

Luzagar:
you can now build with the debug register


caseif:
That use case should be covered by the new `--reghio` flag exposed by the configure script.

caseif:
Oh whoops, I misunderstood what this change was doing. Looking over it again that was definitely a mistake on my part - thanks for catching it!

PR #2991, opened 2025-12-23
f_op debug work

ItsNiklas:
- Work on `f_op` debug matching
- Most of the big files I have not yet worked on, only `f_op_actor_mng`
- Linked some of the TUs for `Shield` and `ShieldD`
- I don't know what I am doing so stuff might be wrong, please check carefully
- For fun I also linked `f_op_msg` for every supported version
- Maybe its too early for the ALL lists, then I will revert it :)


src/f_op/f_op_actor_mng.cpp
```diff
@@ -521,29 +544,36 @@ void fopAcM_setCullSizeSphere(fopAc_ac_c* i_actor, f32 i_minX, f32 i_minY, f32 i
...
-    f32 maxZ = i_actor->scale.z * jointNode->getMax()->z;
+    maxZ = i_actor->scale.z * jointNode->getMax()->z;
+    maxY = i_actor->scale.y * jointNode->getMax()->y;
+    maxX = i_actor->scale.x * jointNode->getMax()->x;
+    minZ = i_actor->scale.z * jointNode->getMin()->z;
+    minY = i_actor->scale.y * jointNode->getMin()->y;
 
-    fopAcM_setCullSizeBox(i_actor, minX, minY, minZ, maxX, maxY, maxZ);
+    fopAcM_setCullSizeBox(i_actor, i_actor->scale.x * jointNode->getMin()->x, minY, minZ, maxX,
+                          maxY, maxZ);
```
ItsNiklas: I don't know right now how to restore `fopAcM_setCullSizeBox2`, debug uses only 5 float registers in this specific order

ItsNiklas:
Also not sure what happened in `dCamera_c::lockonCamera(long)`, no clear difference, the function is too big and has a lot of issues

src/f_op/f_op_actor_mng.cpp
roeming: Because they're called in a close to opposite order, I think there are no variables, all of them are created as a parameter inside the `fopAcM_setCullSizeBox` call

ItsNiklas: Perfect, thank you. Pattern makes sense, yes.

include/d/d_event.h
```diff
@@ -150,7 +150,22 @@ class dEvt_control_c {
...
+        BOOL ret = FALSE;
+        BOOL order_ok = TRUE;
+
+        if (mEventStatus != 0 && mEventStatus != 2) {
+            order_ok = FALSE;
+        }
+        if (order_ok && mDebugStb == 0) {
+            ret = TRUE;
+        }
+        return ret;
```
TakaRikka: does this not match as
```
return (mEventStatus == 0 || mEventStatus == 2) && !mDebugStb;
```
?

include/f_op/f_op_actor_mng.h
```diff
@@ -412,7 +414,7 @@ inline BOOL fopAcM_CULLSIZE_IS_BOX(int i_culltype) {
            i_culltype == fopAc_CULLBOX_CUSTOM_e;
 }
 
-inline const cXyz& fopAcM_getCullSizeSphereCenter(const fopAc_ac_c* i_actor) {
+inline const cXyz& fopAcM_getCullSizeSphereCenter(const fopAc_ac_c* i_actor) { // This should be Vec& but GCN inlines break
```
TakaRikka: i'd write `This should return Vec&` just to be clear

src/d/d_path.cpp
```diff
@@ -63,7 +63,7 @@ dPath* dPath_GetNextRoomPath(dPath const* p_path, int room_no) {
     return &path->m_path[next_id];
 }
 
-u8 dPath_GetPolyRoomPathVec(cBgS_PolyInfo const& poly, cXyz* p_pathVec, int* param_2) {
+int dPath_GetPolyRoomPathVec(cBgS_PolyInfo const& poly, cXyz* p_pathVec, int* param_2) {
```
TakaRikka: is this for sure correct? needing to cast the value everywhere where its used feels fake

src/f_op/f_op_actor_tag.cpp
```diff
@@ -12,8 +12,9 @@ int fopAcTg_ToActorQ(create_tag_class* i_createTag) {
     return cTg_Addition(&g_fopAcTg_Queue, i_createTag);
 }
 
-int fopAcTg_ActorQTo(create_tag_class* i_createTag) {
-    return cTg_SingleCutFromTree(i_createTag);
+void fopAcTg_ActorQTo(create_tag_class* i_createTag) {
+    int _ = cTg_SingleCutFromTree(i_createTag);
+    return;
```
TakaRikka: return can be removed

include/f_op/f_op_actor_mng.h
ItsNiklas: I removed the comment because it still matches debug now. I still think it's the wrong return type.

TakaRikka (approved):

PR #2998, opened 2025-12-25
Npc henna documentation

Luzagar:


include/d/actor/d_a_npc_henna.h
```diff
@@ -71,56 +71,56 @@ class npc_henna_class {
     /* 0x704 */ s16 field_0x704;
     /* 0x706 */ s16 field_0x706;
     /* 0x708 */ u8 field_0x708;
-    /* 0x709 */ u8 field_0x709;
-    /* 0x70A */ u8 field_0x70a;
-    /* 0x70B */ u8 field_0x70b;
+    /* 0x709 */ u8 wait_time;
```
TakaRikka, line 74: in general for all these timer vars, i prefer to name them `timer` instead of `time` to indicate that they actually count down

src/d/actor/d_a_npc_henna.cpp
```diff
@@ -447,223 +447,223 @@ static void henna_ride(npc_henna_class* i_this) {
...
+                               (s16)stick_ang < (s16) - (0x6000 + AREG_S(6)))
                     {
-                        if (i_this->field_0x660 > 0x14 && i_this->mAnmResIndex != 6) {
+                        if (i_this->paddle_time > 0x14 && i_this->res_id != 6) {
                             anm_init(i_this, 6, 10.0f, 2, 1.0f);
                         }
-                        i_this->field_0x660++;
-                        if (i_this->field_0x660 > 0x14) {
+                        i_this->paddle_time++;
+                        if (i_this->paddle_time > 0x14) {
```
TakaRikka, line 567: for any of these timer or counter vars, changing the compared number from hex to decimal would be appreciated

src/d/actor/d_a_npc_henna.cpp
```diff
@@ -848,94 +848,94 @@ static void demo_camera(npc_henna_class* i_this) {
 
     cXyz unkXyz;
     cXyz unkXyz2;
-    cXyz unkXyz3;
-    cXyz unkXyz4;
+    cXyz mae;
+    cXyz ato;
```
TakaRikka, line 852: i think instead of these names, i think they were originally just "work" variables where they were sometimes reused. so i might just call all of these like `work1`, `work2`, etc

src/d/actor/d_a_npc_henna.cpp
```diff
@@ -848,94 +848,94 @@ static void demo_camera(npc_henna_class* i_this) {
...
+    fopAc_ac_c* actor = &i_this->actor;
     fopAc_ac_c* unusedPlayer = (fopAc_ac_c*)dComIfGp_getPlayer(0);
     camera_class* unusedCamera = dComIfGp_getCamera(dComIfGp_getPlayerCameraID(0));
 
     camera_class* camera = (camera_class*)dComIfGp_getCamera(0);
 
-    cXyz target_pos;
-    cXyz lookat_pos;
+    cXyz mae;
+    cXyz ato;
```
TakaRikka, line 924: like here i prob would've left the old names. for reference, "mae" and "ato" are japanese for "before and "after", which i think were used when they were doing mtx multVec operations

src/d/actor/d_a_npc_henna.cpp
```diff
@@ -971,104 +971,104 @@ static void demo_camera_shop(npc_henna_class* i_this) {
...
-    fshop_class* koro_shop;
+    fshop_class* koro2;
     fs_koro2_s* koro;
 
-    cXyz unkXyz_78;
-    cXyz unkXyz_6c;
-    cXyz unkXyz_60;
-    cXyz unkXyz_54;
+    cXyz mae;
+    cXyz ato;
```
TakaRikka, line 988: would also call these work vars

src/d/actor/d_a_npc_henna.cpp
```diff
@@ -848,94 +848,94 @@ static void demo_camera(npc_henna_class* i_this) {
...
         lookat_angle -= target_angle;
-        if (lookat_angle < tolerance && lookat_angle > (s16)-tolerance) {
-            lookat_angle = (s16)-cM_atan2s(lookat_pos.y,
-                                    JMAFastSqrt(lookat_pos.x * lookat_pos.x + lookat_pos.z * lookat_pos.z));
+        if (lookat_angle < range && lookat_angle > (s16)-range) {
+            lookat_angle = (s16)-cM_atan2s(ato.y,
+                                    JMAFastSqrt(ato.x * ato.x + ato.z * ato.z));
             target_angle = (s16)-cM_atan2s(
-                target_pos.y, JMAFastSqrt(target_pos.x * target_pos.x + target_pos.z * target_pos.z));
+                mae.y, JMAFastSqrt(mae.x * mae.x + mae.z * mae.z));
```
TakaRikka, line 936: also for these square operations, you can use the `SQUARE` macro to simplify it a bit, like
`JMAFastSqrt(SQUARE(mae.x) + SQUARE(mae.z))`

include/d/actor/d_a_npc_henna.h
```diff
@@ -71,56 +71,56 @@ class npc_henna_class {
...
-    /* 0x738 */ f32 field_0x738;
+    /* 0x734 */ u8 no_draw;
+    /* 0x738 */ f32 rain_time;
     /* 0x73C */ s16 timer[10];
-    /* 0x750 */ s16 field_0x750;
-    /* 0x752 */ s16 cam_mode;
-    /* 0x754 */ s16 field_0x754;
+    /* 0x750 */ s16 talk_time;
+    /* 0x752 */ s16 demo_mode;
+    /* 0x754 */ s16 demo_camera_no;
```
TakaRikka, line 90: this looks like it should be `demo_timer` from what i can tell

include/d/actor/d_a_npc_henna.h
```diff
@@ -71,56 +71,56 @@ class npc_henna_class {
...
-    /* 0x7B7 */ u8 field_0x7b7;
-    /* 0x7B8 */ u8 field_0x7b8;
+    /* 0x7B6 */ u8 zoom_time;
+    /* 0x7B7 */ u8 talk_ct;
+    /* 0x7B8 */ u8 msg_time;
     /* 0x7B9 */ u8 field_0x7b9;
-    /* 0x7BA */ u8 field_0x7ba;
-    /* 0x7BC */ f32 field_0x7bc;
-    /* 0x7C0 */ f32 field_0x7c0;
+    /* 0x7BA */ u8 demo_time;
```
TakaRikka, line 108: this seems like something less generic than the other demo_timer so it probably needs a more specialized name

include/d/actor/d_a_npc_henna.h
```diff
@@ -71,56 +71,56 @@ class npc_henna_class {
...
+    /* 0x752 */ s16 demo_mode;
+    /* 0x754 */ s16 demo_camera_no;
     /* 0x756 */ u16 field_0x756;
     /* 0x758 */ s16 field_0x758;
     /* 0x75A */ u8 field_0x75a[0x75c - 0x75a];
     /* 0x75C */ s16 field_0x75c;
-    /* 0x760 */ cXyz field_0x760;
-    /* 0x76C */ cXyz field_0x76c;
+    /* 0x760 */ cXyz demo_eye;
+    /* 0x76C */ cXyz demo_way;
```
TakaRikka, line 96: i would probably call this `demo_cam_center` to keep consistent. also demo_eye can be `demo_cam_eye` and so on so forth

src/d/actor/d_a_npc_henna.cpp
```diff
@@ -1847,203 +1848,203 @@ static void demo_camera_shop(npc_henna_class* i_this) {
...
             mDoGph_gInf_c::fadeIn(0.1f, g_blackColor);
         }
 
         dComIfGp_setDoStatusForce(0x22, 0);
 
-        if (unk_bss_5136 == 0) {
+        if (S_stick_x == 0) {
             old_stick_x = 0.0f;
-            unk_bss_5136 = 1;
+            S_stick_x = 1;
```
TakaRikka, line 2036: im pretty sure this pattern is just `static f32 old_stick_x = 0.0f;`? which would automatically generate the bss data here instead of manually defining it

include/d/actor/d_a_npc_henna.h
```diff
@@ -19,47 +19,47 @@ class npc_henna_class {
     /* 0x0   */ fopEn_enemy_c actor;
     /* 0x5AC */ request_of_phase_process_class phase;
     /* 0x5B4 */ u8 arg0;
-    /* 0x5B5 */ s8 field_0x5b5;
+    /* 0x5B5 */ s8 talk_mode;
     /* 0x5B6 */ s8 field_0x5b6;
-    /* 0x5B8 */ s16 mIsTalking;
+    /* 0x5B8 */ s16 talk_check;
```
TakaRikka, line 24: since this only ever seems to be set to 0 or 1, i would keep this as `is_talk`, and make sure its comparisons are updated to reflect its boolean nature by doing things like `!is_talk` instead of `is_talk == 0`

src/d/actor/d_a_npc_henna.cpp
```diff
@@ -2272,90 +2273,90 @@ static void env_control(npc_henna_class* i_this) {
...
     }
 }
 
 static int daNpc_Henna_Execute(npc_henna_class* i_this) {
     fopAc_ac_c* unusedPlayer = dComIfGp_getPlayer(0);
 
-    cXyz lookat_pos;
-    cXyz target_pos;
+    cXyz mae;
+    cXyz ato;
```
TakaRikka, line 2323: these are probably better as before or as work vars

src/d/actor/d_a_npc_henna.cpp
```diff
@@ -2272,90 +2273,90 @@ static void env_control(npc_henna_class* i_this) {
...
-    if (i_this->field_0x70a != 0) {
-        i_this->field_0x70a -= 1;
-        if (i_this->field_0x70b == 0) {
-            i_this->field_0x694 = 1;
-            i_this->field_0x693 = 0;
+    if (i_this->fade_time != 0) {
+        i_this->fade_time -= 1;
+        if (i_this->fade_type == 0) {
+            i_this->hide_body_flag = 1;
+            i_this->rod_draw_flag = 0;
```
TakaRikka, line 2330: all the flag type vars should also be treated like booleans, i.e. `i_this->hide_body_flag = TRUE` and such

src/d/actor/d_a_npc_henna.cpp
```diff
@@ -1847,203 +1848,203 @@ static void demo_camera_shop(npc_henna_class* i_this) {
...
         cMtx_YrotS(*calc_mtx, fshop->field_0x4060);
-        unkXyz_78.x = -500.0f;
-        unkXyz_78.y = 600.0f;
-        unkXyz_78.z = 0.0f;
-        MtxPosition(&unkXyz_78, &unkXyz_60);
-        unkXyz_60 += koro_shop->actor.current.pos;
+        mae.x = -500.0f;
+        mae.y = 600.0f;
+        mae.z = 0.0f;
+        MtxPosition(&mae, &koro2_eye);
```
Luzagar, line 1858: i think mae is approriate here beacause it's use in in a MtxPosition and

src/d/actor/d_a_npc_henna.cpp
```diff
@@ -1704,114 +1705,114 @@ static void demo_camera_shop(npc_henna_class* i_this) {
...
-        if (i_this->field_0x754 == 10) {
+        if (i_this->demo_camera_no == 10) {
             fshop = (fshop_class*)fpcM_Search(s_shop_sub, i_this);
             if (fshop != NULL) {
                 fshop->field_0x4010 = 1;
             }
 
-            unkXyz_6c.set(171.0f, 0.0f, 432.0f);
-            daPy_getPlayerActorClass()->setPlayerPosAndAngle(&unkXyz_6c, 0x4000, 0);
+            ato.set(171.0f, 0.0f, 432.0f);
```
Luzagar, line 1767: also what it motive to name it ato because i'v seen this patern in d_a_b_ds (executeBattle2Dead)  and d_a_b_dr(executeTailHit) and i think if there is mae , there is also  ato ?

src/d/actor/d_a_npc_henna.cpp
Luzagar: there are both used in a MtxPosition in line 2571 so i think there name is approriate ?

src/d/actor/d_a_npc_henna.cpp
Luzagar: and i'v seen other actor where the use the eye in a MtxPosition

src/d/actor/d_a_npc_henna.cpp
Luzagar: it break match in retail

src/d/actor/d_a_npc_henna.cpp
TakaRikka: in the past i used mae ato in some places id disagree with now, but nowadays i try to only use it when its being used explicitly for multvec operations, otherwise i think work is a more appropriate name. since mae and ato imply that theres some sort of operation happening between the two, which is not always the case

src/d/actor/d_a_npc_henna.cpp
roeming: matches and links retail with 
`
        static f32 old_stick_x = 0.0f;
        static f32 old_stick_sx = 0.0f;
`
You just need to remove the static declarations of bss (which are just the static guards) and the old_sticks at the top of the function.

PR #2999, opened 2025-12-25
match d_a_e_mk debug

roeming:


src/d/actor/d_a_e_mk.cpp
```diff
@@ -781,44 +778,49 @@ static void damage_check(e_mk_class* i_this) {
...
+                fpcM_Search(s_obj_delete, i_this);
+                i_this->action = 30;
+                i_this->mode = 0;
+                i_this->sound.startCreatureVoice(Z2SE_EN_MK_V_DEAD, -1);
+                i_this->invulnerabilityTimer = 20000;
+                Z2GetAudioMgr()->subBgmStop();
+
+#if !DEBUG
+                dScnPly_c::nextPauseTimer = 0;
+#endif
```
TakaRikka: is this specifically a debug change? or wii/shield change?

roeming: Looks like it's just a GCN thing, I'll change that

TakaRikka (approved):

PR #3001, opened 2025-12-26
match most of d_a_e_wb debug

roeming:
The only issue was `demo_camera`, not sure if there's something weird where the function should be declared to return something but effectively doesn't.

### DIFF
--- a/src/d/actor/d_a_e_wb.cpp
+++ b/src/d/actor/d_a_e_wb.cpp
@@ -1273,7 +1273,7 @@ static void e_wb_f_run(e_wb_class* i_this) {
 }
 
 static void* s_wbstart_sub(void* i_actor, void* i_data) {
-    (void)i_data;
+    UNUSED(i_data);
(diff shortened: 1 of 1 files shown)


src/d/actor/d_a_e_wb.cpp
```diff
@@ -1260,13 +1273,14 @@ static void e_wb_f_run(e_wb_class* i_this) {
 }
 
 static void* s_wbstart_sub(void* i_actor, void* i_data) {
-    if (fopAcM_IsActor(i_actor) && fopAcM_GetName(i_actor) == PROC_E_WB &&
-        static_cast<e_wb_class*>(i_actor)->field_0x79d == 0)
-    {
-        static_cast<e_wb_class*>(i_actor)->field_0x7a6 = 1;
-        static_cast<e_wb_class*>(i_actor)->field_0x7a7 = 1;
+    (void)i_data;
```
TakaRikka: should be able to use the UNUSED macro for this now

TakaRikka (approved):

PR #3002, opened 2025-12-26
Work on d_cc_d debug

hatal175:


src/SSystem/SComponent/c_m3d.cpp
```diff
@@ -1978,6 +1978,47 @@ void cM3d_CalcVecZAngle(Vec const& param_0, csXyz* param_1) {
     param_1->z = 0;
 }
 
+#ifdef DEBUG
```
TakaRikka, line 1981: are these debug specific? maybe just a version diff?

hatal175: I believe so. They only appear on the shield debug version symbols and they since they appear in the c_m3d TU, I don't see how they'll be stripped otherwise.

hatal175: I believe so. They only appear on the shield debug version symbols and they since they appear in the c_m3d TU, I don't see how they'll be stripped otherwise.

TakaRikka (approved):

PR #3008, opened 2025-12-28
f_op work

ItsNiklas:
- Continuation of #2991
- Tried to check all supported versions for guards
- `fopAcM_entrySolidHeap_` is my mortal enemy
- Maybe someone with more experience needs to give the remaining issues another pass in the future
- `showAssert_f_va` and `setWarningMessage_f_va` I don't quite understand, they are also wrongly in the namespace.
- Linked quite a few versions :)

### DIFF
--- a/include/d/d_error_msg.h
+++ b/include/d/d_error_msg.h
@@ -12,8 +12,10 @@ struct dDvdErrorMsg_c {
     static u8 execute();
 };
 
+#if !PLATFORM_GCN
 struct dConnectErrorMsg_c {
     static void disable();
 };
+#endif
 
 #endif /* D_D_ERROR_MSG_H */
--- a/include/f_op/f_op_actor_map.h
+++ b/include/f_op/f_op_actor_map.h
@@ -1,6 +1,0 @@
-#ifndef F_OP_ACTOR_MAP_H_
-#define F_OP_ACTOR_MAP_H_
-
-
-
-#endif  // F_OP_ACTOR_MAP_H_
--- a/src/d/d_path.cpp
+++ b/src/d/d_path.cpp
@@ -63,7 +63,7 @@ dPath* dPath_GetNextRoomPath(dPath const* p_path, int room_no) {
     return &path->m_path[next_id];
 }
 
-#if !PLATFORM_GCN
+#if !PLATFORM_GCN  // Fakematch due to differing return type on non-GCN platforms
 int dPath_GetPolyRoomPathVec(cBgS_PolyInfo const& poly, cXyz* p_pathVec, int* param_2) {
 #else
 u8 dPath_GetPolyRoomPathVec(cBgS_PolyInfo const& poly, cXyz* p_pathVec, int* param_2) {
--- a/src/f_op/f_op_actor_mng.cpp
+++ b/src/f_op/f_op_actor_mng.cpp
@@ -80,7 +80,7 @@ void l_HIO::genMessage(JORMContext* mctx) {
 }
 
 void l_HIO::listenPropertyEvent(const JORPropertyEvent* property) {
-    UNUSED(property);
+    (void)property;
     JORMContext* mctx = attachJORMContext(8);
     JORReflexible::listenPropertyEvent(property);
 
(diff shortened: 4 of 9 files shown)


include/d/d_error_msg.h
```diff
@@ -12,4 +12,8 @@ struct dDvdErrorMsg_c {
     static u8 execute();
 };
 
+struct dConnectErrorMsg_c {
```
TakaRikka, line 16: i'd wrap this in a platform define since it doesnt apply to GCN

include/f_op/f_op_actor_map.h
```diff
@@ -0,0 +1,6 @@
+#ifndef F_OP_ACTOR_MAP_H_
```
TakaRikka: is this header needed? iirc nothing from actor_map actually stayed in the game, so it can probably be removed

src/d/d_path.cpp
```diff
@@ -63,7 +63,11 @@ dPath* dPath_GetNextRoomPath(dPath const* p_path, int room_no) {
     return &path->m_path[next_id];
 }
 
+#if !PLATFORM_GCN
```
TakaRikka: maybe a fakematch comment here, this seems suspicious

src/f_op/f_op_actor_mng.cpp
```diff
@@ -100,6 +94,35 @@ void l_HIO::listenPropertyEvent(const JORPropertyEvent* property) {
 
     releaseJORMContext(mctx);
 }
+
+void fopAcM_setWarningMessage_f(const fopAc_ac_c* i_actor, const char* i_filename, int i_line,
+                                const char* i_msg, ...) {
+    UNUSED(i_msg);
```
TakaRikka: i_msg isnt actually unused in either of these though? if this is needed for matching, its probably something else and not actually an unused var

src/f_op/f_op_actor_mng.cpp
```diff
@@ -393,102 +452,262 @@ bool fopAcM_entrySolidHeap_(fopAc_ac_c* i_actor, heapCallbackFunc i_heapCallback
...
+                    adjustedHeap = ALIGN_NEXT(heap->getHeapSize() - heap->getFreeSize(), 0x20);
+#if PLATFORM_SHIELD
+                    if (i_size < adjustedHeap + margin || fopAcM::HeapSkipMargin) {
+#elif VERSION == VERSION_WII_USA_R2 || VERSION == VERSION_WII_PAL
+                    if (i_size < adjustedHeap + margin || fopAcM::HeapAdjustUnk) {
+#elif VERSION == VERSION_WII_USA_R0
+                    if (i_size < adjustedHeap + margin) {
+#else
+                    if (i_size < adjustedHeap + margin) {
+#endif
```
TakaRikka: arent these two lines the same?

src/f_op/f_op_actor_mng.cpp
```diff
@@ -1384,39 +1624,54 @@ fpc_ProcID fopAcM_createItem(const cXyz* i_pos, int i_itemNo, int i_itemBitNo, i
...
 
     u8 item_no = check_itemno(i_itemNo);
-    u32 params = MAKE_ITEM_PARAMS(item_no, i_itemBitNo, 0xFF, param_7);
+    int unk = -1;
+    fpc_ProcID ret;
+    int i;
+    
+    u32 params = MAKE_ITEM_PARAMS(item_no, i_itemBitNo, unk, param_7);
+    // u32 params = ((item_no & 0xFF) << 0x0 | (i_itemBitNo & 0x7F) << 0x8 | (unk & 0xFF) << 0x10 | (param_7 & 0xF) << 0x18);
+    // u32 params =  (item_no & 0xFF) << 0x0 | (i_itemBitNo & 0xFF) << 0x8 | (unk & 0xFF) << 0x10 | (param_7 & 0xF) << 0x18;
```
TakaRikka: can these be removed?

src/f_op/f_op_actor_mng.cpp
```diff
@@ -1384,39 +1624,54 @@ fpc_ProcID fopAcM_createItem(const cXyz* i_pos, int i_itemNo, int i_itemBitNo, i
...
 
     switch (i_itemNo) {
     case fpcNm_ITEM_RECOVERY_FAILY:
-        return fopAcM_create(PROC_Obj_Yousei, 0xFFFFFFFF, i_pos, i_roomNo, i_angle, i_scale, -1);
+        ret = fopAcM_create(PROC_Obj_Yousei, 0xFFFFFFFF, i_pos, i_roomNo, i_angle, i_scale, -1);
+        break;
 #if DEBUG
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wsometimes-uninitialized"
```
TakaRikka: idk about putting clang pragmas in code like this. id rather them just added to the clangd file

src/f_op/f_op_draw_iter.cpp
```diff
@@ -34,7 +35,9 @@ create_tag_class* fopDwIt_Begin() {
 }
 
 create_tag_class* fopDwIt_Next(create_tag_class* i_createTag) {
+    UNUSED(&i_createTag);
```
TakaRikka: again this param is actually used, so this seems fake

src/f_op/f_op_msg_mng.cpp
```diff
@@ -156,9 +165,9 @@ fpc_ProcID fopMsgM_messageSet(u32 i_msgIdx, u32 param_1) {
         dComIfGp_isHeapLockFlag() != 1)
     {
         return fpcM_ERROR_PROCESS_ID_e;
-
     }
 
+    fopAc_ac_c* NULL_ = NULL;
```
TakaRikka: i'd probably just call these "actor" or something, since they probably werent meant to be null originally but the functionality was removed

src/f_op/f_op_scene.cpp
```diff
@@ -19,7 +19,8 @@ static int fopScn_Execute(scene_class* i_this) {
 }
 
 static int fopScn_IsDelete(void* i_this) {
-    return fpcMtd_IsDelete((process_method_class*)static_cast<scene_class*>(i_this)->submethod, i_this);
+    int ret = fpcMtd_IsDelete((process_method_class*)static_cast<scene_class*>(i_this)->submethod, i_this);
```
TakaRikka: prefer C casts for `f` TUs

src/f_op/f_op_scene_req.cpp
```diff
@@ -47,25 +47,30 @@ static cPhs__Step fopScnRq_phase_IsDoneOverlap(scene_request_class* i_sceneReq)
 static BOOL l_fopScnRq_IsUsingOfOverlap;
 
 static cPhs__Step fopScnRq_phase_Done(scene_request_class* i_sceneReq) {
+    
     if (i_sceneReq->create_request.parameters != 1) {
-        scene_class* scene = (scene_class*)fpcM_SearchByID(i_sceneReq->create_request.creating_id);
+        scene_class* scene;
+        UNUSED(scene);
```
TakaRikka: this isnt unused, so this doesnt make sense

src/d/d_path.cpp
ItsNiklas: I didn't know how to fix it otherwise. GCN uses `clrlwi. r0, r3, 24` on the return value and every other version uses `cmpwi r3, 0x0`. I doubt it was casted everywhere (5 uses).

src/f_op/f_op_actor_mng.cpp
ItsNiklas: Yeah its used but it solves regalloc on Debug. I will just use `(void)` directly.

PR #3009, opened 2025-12-29
m_Do_ext retail linked + debug improvements

caseif:
This PR fixes weak function order for `m_Do_ext` and improves debug matching.

### DIFF
--- a/include/JSystem/JKernel/JKRHeap.h
+++ b/include/JSystem/JKernel/JKRHeap.h
@@ -36,6 +36,7 @@ public:
     };
 
 public:
+    JKRHeap(u32 size, JKRHeap* parent, bool errorFlag);
     JKRHeap(void* data, u32 size, JKRHeap* parent, bool errorFlag);
     virtual ~JKRHeap();
 
@@ -112,6 +113,7 @@ public:
     void lock() { OSLockMutex(&mMutex); }
     void unlock() { OSUnlockMutex(&mMutex); }
     u32 getHeapSize() { return mSize; }
+    u8 getCurrentGroupId() { return 0; }
 
 protected:
     /* 0x00 */  // vtable
@@ -151,9 +153,7 @@ public:
     static u32 getMemorySize(void) { return mMemorySize; }
     static JKRHeap* getRootHeap() { return sRootHeap; }
 
-#if PLATFORM_WII || PLATFORM_SHIELD
     static JKRHeap* getRootHeap2() { return sRootHeap2; }
-#endif
 
     static JKRHeap* getSystemHeap() { return sSystemHeap; }
     static JKRHeap* getCurrentHeap() { return sCurrentHeap; }
@@ -176,9 +176,7 @@ public:
 
     static JKRHeap* sRootHeap;
 
-#if PLATFORM_WII || PLATFORM_SHIELD
     static JKRHeap* sRootHeap2;
-#endif
 
     static JKRHeap* sSystemHeap;
     static JKRHeap* sCurrentHeap;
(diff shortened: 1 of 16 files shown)


include/JSystem/JKernel/JKRHeap.h
```diff
@@ -151,9 +153,7 @@ class JKRHeap : public JKRDisposer {
     static u32 getMemorySize(void) { return mMemorySize; }
     static JKRHeap* getRootHeap() { return sRootHeap; }
 
-#if PLATFORM_WII || PLATFORM_SHIELD
     static JKRHeap* getRootHeap2() { return sRootHeap2; }
-#endif
```
TakaRikka, line 156: is there a reason this stuff and the mDoExt stuff had the platform defines removed? it doesnt seem like these existed in GCN

src/m_Do/m_Do_ext.cpp
```diff
@@ -108,18 +124,14 @@ int mDoExt_baseAnm::play() {
 
 int mDoExt_bpkAnm::init(J3DMaterialTable* i_matTable, J3DAnmColor* i_bpk, int i_anmPlay,
                         int i_attribute, f32 i_rate, s16 i_startF, s16 i_endF) {
+    JUT_ASSERT(371, (i_anmPlay == 0 || getFrameCtrl() != 0 || isCurrentSolidHeap()) && i_matTable != NULL && i_bpk != NULL);
```
TakaRikka: think you can use FALSE and NULL here without it messing up the string

src/m_Do/m_Do_ext.cpp
```diff
@@ -655,71 +688,178 @@ JKRExpHeap* mDoExt_getGameHeap() {
...
+        mDoExt_getGameHeap()->changeGroupID(7);
+    }
+}
+
+intptr_t mDoExt_getSafeCommandHeapSize() {
+    return safeCommandHeapSize;
+}
+
 JKRExpHeap* archiveHeap;
+intptr_t safeArchiveHeapSize = -1;
```
TakaRikka, line 775: this seems like its just a size, so probably shouldnt be intptr_t

src/m_Do/m_Do_ext.cpp
```diff
@@ -3493,8 +3707,13 @@ J3DModel* mDoExt_J3DModel__create(J3DModelData* i_modelData, u32 i_modelFlag, u3
         J3DModel* model = new J3DModel();
 
         if (model != NULL) {
+#if DEBUG
+            model->getBaseTRMtx()[0][0] = MAXFLOAT;
+#endif
```
TakaRikka, line 3724: debug or version diff?

include/JSystem/JKernel/JKRHeap.h
caseif: The guarded symbols should be stripped out by the linker and not end up in the final DOL, so my thinking was that it would be better to simplify things and remove the redundant version checks. I don't think any are present in the Wii maps either and only show up in ShieldD.

src/m_Do/m_Do_ext.cpp
caseif: `intptr_t` would be more portable in this case, although I agree it's not really appropriate here. Maybe we could define `ssize_t` instead?

src/m_Do/m_Do_ext.cpp
caseif: This appears to be debug-specific.

caseif:
Resolved, a symbol name was accidentally removed.

PR #3011, opened 2025-12-31
d_a_e_rdy debug work

roeming:
Almost all functions match in debug, with the only one left being `demo_camera`, which has bad regalloc possibly from functions with the wrong return type. The biggest change was for `MtxPull` to return `void` instead of `Mtx*`. The previous version caused lots of bad regalloc for `demo_camera`, and I found this drastically improved matching with no regressions.

### DIFF
--- a/src/d/actor/d_a_e_rdy.cpp
+++ b/src/d/actor/d_a_e_rdy.cpp
@@ -683,11 +683,7 @@ static BOOL way_check(e_rdy_class* i_this) {
             dBgS_LinChk lin_chk;
             lin_chk.Set(&start, &end, _this);
             if (dComIfG_Bgsp().LineCross(&lin_chk)) {
-#if PLATFORM_WII || PLATFORM_GCN
-                angle += 0x1000;
-#else
-                angle += (s16)0x1000;
-#endif
+                ADD_ANGLE(angle, 0x1000);
             } else {
                 i_this->mTargetAngleY = angle;
                 return TRUE;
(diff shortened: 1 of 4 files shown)


src/d/actor/d_a_e_rdy.cpp
```diff
@@ -1124,7 +1169,11 @@ static void e_rdy_bow_run(e_rdy_class* i_this) {
 
     case 1:
         target_speed = run_speed;
+#if PLATFORM_GCN || PLATFORM_WII
         target_angle += 0x8000;
+#else
+        target_angle += (s16)0x8000;
+#endif
```
TakaRikka: is there any way to rectify these for all versions? i feel like its possible there was a macro involved for angle manipulation

src/d/actor/d_a_e_rdy.cpp
```diff
@@ -2061,7 +2122,10 @@ static BOOL body_gake(e_rdy_class* i_this) {
 
 static void e_rdy_damage(e_rdy_class* i_this) {
     fopEn_enemy_c* a_this = &i_this->actor;
+    fopEn_enemy_c* a_this2 = a_this;
```
TakaRikka: im guessing that one of these is the base actor class, and the other is the enemy class, or something like that

roeming: That works, however this is the way it works out
```
fopAc_ac_c* a_this = &i_this->actor;
fopEn_enemy_c* e_this = (fopEn_enemy_c*)a_this;
```
It's a little funny, but does match if you're ok with that. My guess is they started every function by retrieving a pointer to the actor, and then additionally needed the enemy subtype.

src/d/actor/d_a_e_rdy.cpp
roeming: In this case, the casting is needed for the Shield and ShieldD versions, and adding a cast doesn't seem to match on Wii or GC.

roeming: I could add a macro for `#define SDEG_180 (s16)0x8000` and have it differ based on version, but there are examples in GC and Wii that still require the cast. So I don't know if it was a macro.

YunataSavior: @roeming and @TakaRikka we have similar angle+casting weirdness in `d_a_npc_ks.cpp`. We should probably create an `angle_vals.h` common header and put the angles in there.

TakaRikka: i dont really want to do a define for specific angles, id prefer just a macro that handles it in a generalized manner

PR #3012, opened 2026-01-02
JKernel debug

Jcw87:


src/JSystem/JKernel/JKRArchivePub.cpp
```diff
@@ -14,11 +14,9 @@ JKRArchive* JKRArchive::check_mount_already(s32 entryNum, JKRHeap* heap) {
         heap = JKRGetCurrentHeap();
     }
 
-    JSUList<JKRFileLoader>& volumeList = getVolumeList();
-    JSUListIterator<JKRFileLoader> iterator;
-    for (iterator = volumeList.getFirst(); iterator != volumeList.getEnd(); ++iterator) {
+    for (JSUListIterator<JKRFileLoader> iterator = sVolumeList.getFirst(); iterator != sVolumeList.getEnd(); ++iterator) {
         if (iterator->getVolumeType() == 'RARC') {
-            JKRArchive* archive = (JKRArchive*)iterator.getObject();
+            JKRArchive* archive = (JKRArchive*)iterator.operator->();
```
TakaRikka, line 19: is this the only way to get this to work? explicit operator calls seems kinda fake-y

Jcw87: image
The only ways I know to force a high register allocation on debug is using a local variable and using a ternary. I think there are more ways to do it, but I don't know how they work.

TakaRikka (approved):

PR #3014, opened 2026-01-03
Misc debug include/class order improvements + link remaining Z2AudioLib TUs

caseif:
This PR includes some miscellaneous improvements to include and class ordering in debug (mainly based on RTTI and assert message order) and also links `Z2AudioMgr` and `Z2SceneMgr` for retail. There are still some lingering orders with include order - for instance, RTTI in `J3DShape` implies `J3DShape` should come before `J3DSys` whereas `J3DPacket` implies the opposite. In any event though, RTTI and string data should generally match better than before.


LagoLunatic:
If two of the Z2 TUs include dolzel.pch, shouldn't all the other Z2 TUs include it too for consistency? Or does that break the order in other TUs?

caseif:
@LagoLunatic It works for most TUs, but `Z2AudioMgr` and `Z2SceneMgr` both place `Z2Calc::cNullVec` in `.rodata` indicating that they either don't use a PCH or use a PCH that doesn't include `Z2Calc.h`. In the latter case presumably it would be a PCH specific to Z2AudioLib so it would be really weird if it skipped a specific Z2 header, so I'm not really sure what the original configuration was. IMO the most plausible scenario I can think of is that it simply wasn't applied consistently, so I've only added it to the TUs that actually benefit.

caseif:
Conflicts are all resolved now.

TakaRikka (approved):

PR #3015, opened 2026-01-04
f_pc mostly linked for Wii/Shield

caseif:
This PR includes various minor improvements for `f_pc` with most of the TUs now being able to be linked for the Wii/Shield versions.


src/f_pc/f_pc_manager.cpp
```diff
@@ -83,47 +90,20 @@ void fpcM_Management(fpcM_ManagementFunc i_preExecuteFn, fpcM_ManagementFunc i_p
...
             Z2GetSoundMgr()->pauseAllGameSound(true);
-            mDoCPd_c::stopMotorWaveHard(0);
+            for (u32 i = 0;
+#if PLATFORM_GCN
+                i < 1;
+#elif PLATFORM_SHIELD && !DEBUG
+                i < 0;
+#else
+                i < 4;
+#endif
```
TakaRikka, line 102: this seems like it should have a value define somewhere. it'll look a lot cleaner

src/f_pc/f_pc_base.cpp
```diff
@@ -112,58 +112,4 @@ int fpcBs_Delete(base_process_class* i_proc) {
...
-    case cPhs_UNK3_e:
-        i_proc->state.create_phase = cPhs_UNK3_e;
-        return cPhs_UNK3_e;
-    case cPhs_ERROR_e:
-    default:
-        i_proc->state.create_phase = cPhs_ERROR_e;
-        return cPhs_ERROR_e;
-    }
-}
+#include "f_pc/f_pc_base_create.inc"
```
TakaRikka: is there a reason for this new include?

src/f_pc/f_pc_manager2.inc
LagoLunatic, line 1: Are you sure this file is real? I assume it's to fix weak order for debug, but debug weak order for this TU can also be fixed with `-sym off` in configure.py. `-sym off` for a debug feels strange and possibly fake, but it seems a bit more likely than a .inc file in f_pc code, as .inc files are usually limited to dolzel code only.

LagoLunatic: Also, in case you weren't aware, f_pc also exists for TWW, and since TWW's maps are complete we know for a fact there was only one .text section for f_pc_manager. This means that there was definitely no inc file for this TU in TWW, so either they added it for TP for some reason (not likely), or they changed the compiler flags for debug.

caseif: If that's the case in TWW's maps then I agree that this is probably fake. I added since it was the only way I could find to get the functions in it to be placed at the bottom of `.text`, but it still doesn't fully match anyway so I can just revert it.

src/f_pc/f_pc_base.cpp
caseif: The create functions are placed at the bottom of `.text` for some reason in ShieldD, but Lago pointed out that this isn't supported by TWW's maps so I'm going to just revert it.

TakaRikka (approved):

PR #3021, opened 2026-01-07
first round of constants cleanup

roeming:
I went through and found a lot of the places where there are float literals which are rational numbers that can be derived, and attempted to derive them. I also went and found lots of places that use the short angle and convert to degrees, and changed to using the macro. In the future, I'm going to look to replace all uses of `x += (s16)(y);`. I also renamed the `ADD_ANGLE` macro to `ADD_ANGLE2`, mostly because there are going to be 2 versions of the macro. The non-numbered version of `ADD_ANGLE` is the version that is consistent across all versions, and `ADD_ANGLE2` is specific to versions that need the cast in shield, but don't match on GC or WII with the cast. I'm open to renaming the macro.

### DIFF
--- a/include/SSystem/SComponent/c_angle.h
+++ b/include/SSystem/SComponent/c_angle.h
@@ -8,11 +8,11 @@
 
 // There are some angles that weren't sign-extended until the shield version
 #if !PLATFORM_SHIELD
-#define ADD_ANGLE_2(x, y) ((x) += (y))
-#define SUB_ANGLE_2(x, y) ((x) -= (y))
+    #define ADD_ANGLE_2(x, y) ((x) += (y))
+    #define SUB_ANGLE_2(x, y) ((x) -= (y))
 #else
-#define ADD_ANGLE_2(x, y) ADD_ANGLE(x, y)
-#define SUB_ANGLE_2(x, y) SUB_ANGLE(x, y)
+    #define ADD_ANGLE_2(x, y) ADD_ANGLE(x, y)
+    #define SUB_ANGLE_2(x, y) SUB_ANGLE(x, y)
 #endif
 
 #define DEG2S_CONSTANT (0x8000 / 180.0f)


src/JSystem/JStudio/JStudio/jstudio-math.cpp
```diff
@@ -62,7 +63,7 @@ void JStudio::math::getFromTransformation_SRxyzT(Vec* param_1, Vec* param_2, Vec
                    param_4[1][0] * dVar9, param_4[1][1] * dVar8, param_4[1][2] * dVar7,
                    param_4[2][0] * dVar9, param_4[2][1] * dVar8, param_4[2][2] * dVar7);
     aTStack_88.getEulerXYZ(&local_a0);
-    local_a0 *= 57.29577951308232;
+    local_a0 *= 180.0 / m_PI_D;
```
LagoLunatic, line 65: JSystem and SSystem are separate libraries. Instead of making JSystem depend on SSystem, maybe `m_PI_D` should go in a JSystem header? It looks like `m_PI_D` is only used in JSystem anyway, so there doesn't need to be a duplicate in SSystem too.

include/JSystem/JParticle/JPAEmitter.h
```diff
@@ -151,7 +151,11 @@ class JPABaseEmitter {
     void setAwayFromAxisSpeed(f32 i_speed) { mAwayFromAxisSpeed = i_speed; }
     void setSpread(f32 i_spread) { mSpread = i_spread; }
     void setLocalTranslation(const JGeometry::TVec3<f32>& i_trans) { mLocalTrs.set(i_trans); }
-    void setLocalRotation(const JGeometry::TVec3<s16>& i_rot) { mLocalRot.set(i_rot.x * 0.005493248f, i_rot.y * 0.005493248f, i_rot.z * 0.005493248f); }
+    void setLocalRotation(const JGeometry::TVec3<s16>& i_rot) {
+        // wrong conversion factor, bug fix should used S2DEG(x)
+        mLocalRot.set(i_rot.x * (360.0f / 0xffff), i_rot.y * (360.0f / 0xffff),
+                      i_rot.z * (360.0f / 0xffff));
```
LagoLunatic: Same here, I think the reason the conversion factor here is slightly different from the one used by SSystem is because JSystem probably had their own unrelated angle defines. Maybe this is less a bug and more a very minor 0.00001% incompatibility between the two libraries?

roeming: Make sense, it looks like the `mLocalRot` is never used anyway? So it's not a huge loss

PR #3029, opened 2026-01-11
e_rd, npc_ks dbg work

YunataSavior:


src/d/actor/d_a_e_rd.cpp
```diff
@@ -25,6 +25,13 @@
 #include "Z2AudioLib/Z2Instances.h"
 #include "f_op/f_op_actor_enemy.h"
 
+// potential fakematch?
+#if PLATFORM_SHIELD
+    #define E_RD_UNSET_FLAG(var, flag) (var) &= (u16)~(flag)
+#else
+    #define E_RD_UNSET_FLAG(var, flag) (var) &= ~(flag)
+#endif
```
TakaRikka: i think this would be better in global.h with a generic name. ive seen this issue in other places like fopEn_enemy_c

src/d/actor/d_a_e_rd.cpp
```diff
@@ -223,30 +229,85 @@ daE_RD_HIO_c::daE_RD_HIO_c() {
...
-    fVar4 = actor->current.pos.z - a_this->current.pos.z;
-    if (fVar1 * fVar1 + fVar3 * fVar3 > fVar2 * fVar2 + fVar4 * fVar4) {
+
+    f32 link_rd_x_diff, coach_rd_x_diff, link_rd_z_diff, coach_rd_z_diff;
+    link_rd_x_diff = player->current.pos.x - a_this->current.pos.x;
+    link_rd_z_diff = player->current.pos.z - a_this->current.pos.z;
+    coach_rd_x_diff = actor->current.pos.x - a_this->current.pos.x;
+    coach_rd_z_diff = actor->current.pos.z - a_this->current.pos.z;
+    // If the XZ-dist of Link is further than the coach, then focus the coach.
+    if (link_rd_x_diff * link_rd_x_diff + link_rd_z_diff * link_rd_z_diff
```
TakaRikka: SQUARE macro can simplify this a bit

src/d/actor/d_a_e_rd.cpp
```diff
@@ -611,16 +673,81 @@ static void* s_b_sub(void* i_actor, void* i_data) {
...
         target_info[i] = NULL;
     }
 
+    // For whatever reason, this function call is missing.
+    // Not sure if it's intentional or not.
+    // In any case, because target_info_count is left at 0 and only incremented
+    // in s_b_sub, the entire rest of this function disappears in release asm.
+#if 0
+    fpcM_Search(s_b_sub, i_this);
+#endif
```
TakaRikka, line 686: is this just an educated guess or does it affect matching? if its a guess i'd probably note it as such just so its clear

src/d/actor/d_a_e_rd.cpp
```diff
@@ -611,16 +673,81 @@ static void* s_b_sub(void* i_actor, void* i_data) {
...
+    if (target_info_count != 0) {
+        cXyz vec1, vec2, vec3;
+        for (int i = 0; i < target_info_count;) {
+            bomb = (dBomb_c*)target_info[i];
+            vec1.x = bomb->current.pos.x - a_this->eyePos.x;
+            vec1.y = 50.0f + bomb->current.pos.y - a_this->eyePos.y;
+            vec1.z = bomb->current.pos.z - a_this->eyePos.z;
+            vec2.x = bomb->current.pos.x - a_this->current.pos.x;
+            vec2.z = bomb->current.pos.z - a_this->current.pos.z;
+            f32 dist1 = JMAFastSqrt(vec1.x * vec1.x + vec1.z * vec1.z);
```
TakaRikka: SQUARE macros again

src/d/actor/d_a_e_rd.cpp
YunataSavior: It's an educated guess that this was ommitted, either by mistake or intentionally. I wrote the above comment to denote that, but IDK if my comment was clear enough.

PR #3033, opened 2026-01-11
Work on c_damagereaction and f_ap_game debug

hatal175:


hatal175:
Not sure about the regression there, it's the only place it affects where it doesn't inline a function it inlined before. Debug is very clear about what setTVec3f should be though.

YunataSavior:
I am still not convinced we've found all the correct compiler settings/versions just yet. I wouldn't worry about it currently

TakaRikka (approved):

PR #3044, opened 2026-01-13
Use attention enums everywhere instead of integer literals

caseif:
This PR fixes a bunch of instances of integer literals being used for attention types/flags instead of the corresponding enum values. This helps a bit with non-GCN versions where the enum values are different. I tried to find as many occurrences of this as I could, although it's likely that there are still some I missed.


caseif:
I have no idea what's going on with the reported `.data`/`.bss` regressions; they don't seem to line up with my local copy.

PR #3051, opened 2026-01-17
npc_besu, npc_toby DBG equiv

YunataSavior:


src/d/d_com_static.cpp
```diff
@@ -204,8 +204,8 @@ int daNpcMsg_setEvtNum(u8 iEvtNum) {
     return 1;
 }
 
-u8 daNpcKakashi_chkSwdTutorialStage() {
-    return strcmp(dComIfGp_getStartStageName(), "F_SP103") == 0 && dComIfG_play_c::getLayerNo(0) == 8;
+bool daNpcKakashi_chkSwdTutorialStage() {
+    return bool(strcmp(dComIfGp_getStartStageName(), "F_SP103") == 0 && dComIfG_play_c::getLayerNo(0) == 8);
```
TakaRikka, line 208: is this the only way to get this to work? casts to bool seem really fake

YunataSavior: I tried various things. I'm convinced they actually did this. The return type MUST be bool, and the only way I could match this function with this return type was via this.

TakaRikka (approved):

PR #3053, opened 2026-01-19
dbg npc_aru (-1 func), npc_maro, and npc_bans equiv

YunataSavior:
* daNpc_Aru_c::adjustMoveDir doesn't match. I don't know how to fix it without breaking GCN

### DIFF
--- a/src/d/actor/d_a_npc_bans.cpp
+++ b/src/d/actor/d_a_npc_bans.cpp
@@ -137,6 +137,51 @@ enum Type {
     /* 0x5 */ TYPE_DEFAULT,
 };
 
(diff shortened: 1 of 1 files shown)


include/d/actor/d_a_npc_maro.h
```diff
@@ -7,7 +7,7 @@
 
 struct daNpc_Maro_HIOParam {
     /* 0x00 */ daNpcT_HIOParam common;
-    /* 0x8C */ s16 field_0x8c;
+    /* 0x8C */ s16 choccai_interval;
```
TakaRikka: i'd rename this either `chokkai_interval` or `tease_interval` personally

PR #3055, opened 2026-01-19
Add debug-specific weak rodata to PCH

caseif:
This PR adds a 0x14-byte `.rodata` object to for debug TUs which use `...rodata` pooling. I'm not sure where this comes from, but pulling it in this way doesn't appear to cause any regressions so I think this is the best solution for the moment.

### DIFF
--- a/include/d/dolzel_base.pch
+++ b/include/d/dolzel_base.pch
@@ -6,6 +6,7 @@
 
 // Fixes weak .data
 #include <cmath> // IWYU pragma: export
+#include <limits> // IWYU pragma: export
 #include "JSystem/J3DGraphBase/J3DMatBlock.h" // IWYU pragma: export
 #include "Z2AudioLib/Z2Calc.h" // IWYU pragma: export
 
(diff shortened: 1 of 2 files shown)


include/weak_rodata_debug.h
```diff
@@ -0,0 +1,12 @@
+#ifndef _WEAK_RODATA_50_H
+#define _WEAK_RODATA_50_H
+
+#include "dolphin/types.h"
+
+static const u8 weak_rodata_debug[0x14] ATTRIBUTE_ALIGN(8) = {
+    0x7f, 0xf7, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+    0x7f, 0xf7, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+    0x7f, 0xbf, 0xff, 0xff,
```
LagoLunatic: Not sure where they come from either but these resemble signaling NaNs just FYI. Two f64 signaling NaNs and one f32 signaling NaN.

caseif: Good catch! In that case it seems likely they come from `limits`, although I'm not sure why the DP NaN shows up twice. I'll update the PR acccordingly though.

TakaRikka (approved):

PR #3057, opened 2026-01-19
Use default parameter values in JORMContext calls where appropriate

caseif:
This PR simplifies calls to `JORMContext` functions by removing explicit parameters which are the same as the default values for the respective functions.


YunataSavior:
@hatal175 @Pheenoh @TakaRikka let's try to get this PR in first and have all other TUs (including my own) rebase and update their calls to JORMContext.

PR #3061, opened 2026-01-20
JUtility wii

Jcw87:


src/JSystem/JUtility/JUTException.cpp
```diff
@@ -10,8 +10,23 @@
 #include <dolphin/dolphin.h>
 #include <stdint.h>
 
+struct CallbackObject {
+    /* 0x00 */ JUTExceptionUserCallback callback;
+    /* 0x04 */ u16 error;
+    /* 0x06 */ u16 pad_0x06;
```
TakaRikka: explicit padding should be able to be removed

src/JSystem/JUtility/JUTGamePad.cpp
```diff
@@ -109,6 +110,11 @@ u32 JUTGamePad::read() {
                 reset_mask |= bittest;
             }
         } else {
+            /*
+            if (!lbl_8074CFA4) {
+                OS_REPORT("game pad read error (%d)\n", mPadStatus[i].err);
+            }
+            */
```
TakaRikka: why commented out?

Jcw87: lbl_8074CFA4 does not already exist in the file, I don't have an official name for it (not even from Wind Waker), and I did not feel like adding it at this time. I can add it if you really want me to.

src/JSystem/JUtility/JUTException.cpp
Jcw87: This was just a copy/paste from another spot in the file, but I'll clean it up a little.

src/JSystem/JUtility/JUTGamePad.cpp
TakaRikka: putting a note about that there at least would prob be good

PR #3139, opened 2026-03-29
JAudio naming pass

PJB3005:


libs/JSystem/src/JAudio2/JASBNKParser.cpp
```diff
@@ -23,7 +23,7 @@ JASBasicBank* JASBNKParser::createBasicBank(void const* stream, JKRHeap* heap) {
     u32 free_size = heap->getFreeSize();
     
     TFileHeader* filep = (TFileHeader*)stream;
-    JUT_ASSERT(59, filep->id == 'IBNK');
+    JUT_ASSERT(59, filep->mMagic == 'IBNK');
```
TakaRikka: we can't be changing names in asserts because it's necessary to match data. this change should be reverted, and can just make a comment in the struct to clarify that it's the magic

libs/JSystem/src/JAudio2/JASDSPInterface.cpp
```diff
@@ -626,21 +625,20 @@ void JASDsp::TChannel::setPitch(u16 param_0) {
...
 
-void JASDsp::TChannel::setMixerVolume(u8 param_0, s16 param_1) {
-    u16* tmp;
+void JASDsp::TChannel::setMixerVolume(u8 outputChannel, s16 volume) {
     JUT_ASSERT(841, dspMutex);
     if (mForcedStop == 0) {
-        tmp = field_0x010[param_0];
-        tmp[1] = param_1;
-        tmp[3] &= 0xff;
+        OutputChannelConfig& cfg = mOutputChannels[outputChannel];
```
TakaRikka: probably want to change this back to a pointer and put the var declaration outside the block. its probably the source of the regression here

TakaRikka (approved):

PR #3146, opened 2026-04-09
PuzzleInfo reconstruction 

Luzagar:
get those name from tww decomp

### DIFF
--- a/include/d/d_save.h
+++ b/include/d/d_save.h
@@ -484,8 +484,8 @@ public:
     char* getHorseName() const { return const_cast<char*>(mHorseName); }
     void setHorseName(const char* i_name) { strcpy(mHorseName, i_name); }
     u8 getClearCount() const { return mClearCount; }
-    u8 getPuzzleInfo() { return mPuzzleData[5]; }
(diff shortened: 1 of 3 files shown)


include/d/d_save.h
```diff
@@ -494,7 +498,7 @@ class dSv_player_info_c {
     /* 0x25 */ char mHorseName[16];
     /* 0x35 */ u8 unk53;
     /* 0x36 */ u8 mClearCount;
-    /* 0x37 */ u8 unk55[5];
+    /* 0x37 */ u8 mPuzzleData[5];
```
LagoLunatic, line 501: Are we sure that this field is a leftover of `mPuzzleData` from TWW? It's a different length and in a different position within the struct. Could be a completely different field, no?
https://github.com/zeldaret/tww/blob/fe4ca85c7988b8206f225849d42de9edbff5fd20/include/d/d_save.h#L453-L455

Luzagar: My reasoning was that they are close to each other, and also it's the only field with a loop initialization.  tp debug, only read  ClearCount, which was used for obj_apzl  in tww  but yeah, that's a guess.

PR #3151, opened 2026-04-11
Fix daE_FB_c::mKeys array size on PC

Pheenoh:


TakaRikka:
no idea why this causes a j2d regression, guessing its something transient so ignoring

TakaRikka (approved):

PR #3157, opened 2026-04-24
Wii PAL d_s_logo improvements

PJB3005:


PJB3005:
report seems to not have regenerated after last push. I think I fixed the regressions though.

TakaRikka (approved):

PR #3158, opened 2026-04-28
d_meter_HIO Wii/debug mostly matching

caseif:
This PR mostly matches `d_meter_HIO` for all Wii versions + ShieldD.


caseif:
The `.data` and `.sdata` regressions don't appear to be real.

LagoLunatic:
> The `.data` and `.sdata` regressions don't appear to be real.

The regressions are 'real' in the sense that you changed the splits to increase the amount of data in that TU, which caused the percentages to go down because the total amount of data increased without increasing the amount of matching data.

If you want objdiff to notice that the data matches to make the percentage go up, you would have to change the symbol names as well (since `lbl_80539340` and `@60962` don't get paired up because they have different name formats).

caseif:
@LagoLunatic Oh I see, I had overlooked that aspect. I've added dummy symbol names for those symbols just for the sake of cleaning up the report.

TakaRikka (approved):

PR #3163, opened 2026-05-19
Name mSideStepLandComboTimer

linkRace:
This is the field used to count the frames to convert a second jump press into a side roll combo

Found this when making https://github.com/TwilitRealm/dusklight/pull/1512 to ensure jump prompts worked as part of the roll override and figured it needed a better name than the (I assume) auto-gen'd decomp value. Let me know if it needs any work

### DIFF
--- a/include/d/actor/d_a_alink.h
+++ b/include/d/actor/d_a_alink.h
@@ -4158,7 +4158,7 @@ public:
     /* 0x02FAD */ u8 mPeepExitID;
     /* 0x02FAE */ u8 mUseButtonFlags;
     /* 0x02FAF */ u8 field_0x2faf;
-    /* 0x02FB0 */ u8 field_0x2fb0;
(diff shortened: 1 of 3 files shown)


include/d/actor/d_a_alink.h
```diff
@@ -4158,7 +4158,7 @@ class daAlink_c : public daPy_py_c {
     /* 0x02FAD */ u8 mPeepExitID;
     /* 0x02FAE */ u8 mUseButtonFlags;
     /* 0x02FAF */ u8 field_0x2faf;
-    /* 0x02FB0 */ u8 field_0x2fb0;
+    /* 0x02FB0 */ u8 mSideStepLandComboTimer;
```
TakaRikka, line 4161: think i'd prefer to call this something more like `mSideRollCheckTimer`

PR #3165, opened 2026-05-28
m_Do_main mostly matching for Wii versions

caseif:
This PR mostly matches `m_Do_main` for all Wii versions. The only remaining mismatches are `crclr cr1eq` instructions related to varargs calls. These can be fakematched for most versions by moving the vararg out to a variable, but this obviously breaks debug. I was able to link the file for `RZDE01_02` with these fakematches, and presumably at least `RZDJ01` and `RZDP01` should also be linkable. `RZDE01_00` is a little trickier to fakematch, so I haven't managed to link that version at all.

I've added a new PCH called `dolzel_minimal` which currently only pulls in `JKRHeap.h`. This allows the fake `JKRHeap::dump_sort` definition to be removed by fixing function ordering. I suspect this new PCH might be real in some capacity - the standard one won't work because the TU uses data pooling but `Z2Calc::cNullVec` and company aren't present, but at the same time I'm not aware of any other way to get MWCC to emit the `JKRHeap` function after `__sinit`.


caseif:
The regression in `ShieldD` is due to `signaling_NaN` constants getting pulled in by the PCH (as they should), but with the wrong placement in `.sdata2`.

LagoLunatic:
> I've added a new PCH called `dolzel_minimal` which currently only pulls in `JKRHeap.h`. This allows the fake `JKRHeap::dump_sort` definition to be removed by fixing function ordering. I suspect this new PCH might be real in some capacity - the standard one won't work because the TU uses data pooling but `Z2Calc::cNullVec` and company aren't present, but at the same time I'm not aware of any other way to get MWCC to emit the `JKRHeap` function after `__sinit`.

Because we know the `d_` TUs used a PCH called `dolzel.pch`, I wonder if the `m_Do_` TUs used a separate PCH called `machine.pch`? Maybe try using the new PCH in every `m_Do_` TU for consistency, and if that doesn't break any of them rename the PCH to make it more clear that this one isn't related to dolzel.

caseif:
@LagoLunatic That's a good call, I've updated the branch with the change implemented. It doesn't seem to cause any regressions (after a small include change in `m_Do_ext`), so I think a machine-specific PCH is very likely to be real.

include/m_Do/machine.h
```diff
@@ -0,0 +1,10 @@
+#ifndef DOLZEL_MINIMAL_H
```
TakaRikka: probably want to update the header guards to reflect the name change

include/m_Do/machine.pch
```diff
@@ -0,0 +1,15 @@
+#ifndef DOLZEL_MINIMAL_PCH
```
TakaRikka, line 1: same here

src/m_Do/m_Do_ext.cpp
```diff
@@ -3,7 +3,8 @@
  * Model, Animation, and Heap Functions
  */
 
-#include "d/dolzel.h" // IWYU pragma: keep
+#include "m_Do/machine.h" // IWYU pragma: keep
+//#include "d/dolzel.h" // IWYU pragma: keep
```
TakaRikka: can be removed?

caseif:
All noted issues should be fixed and the branch has been rebased on latest main.

TakaRikka (approved):

